PR #3, opened 2016-03-08
Use Symbol to represent key names

philwills:


philwills:
cc @LATaylor-guardian 

I'm not sure whether this is a good idea or not and would value any opinions.  

I quite like the distinction it provides between keys and values.

src/main/scala/com/gu/scanamo/ScanamoRequest.scala
```diff
@@ -41,22 +41,22 @@ object ScanamoRequest {
...
-      .withExpressionAttributeValues(asAVMap(s":$k" -> v))
+      .withKeyConditionExpression(s"#K = :${k.name}")
+      .withExpressionAttributeNames(Map("#K" -> k.name).asJava)
+      .withExpressionAttributeValues(asAVMap(Symbol(s":${k.name}") -> v))
   }
 
-  private def asAVMap[K](kvs: (String, K)*)(implicit fk: DynamoFormat[K]) =
-    Map(kvs: _*).mapValues(fk.write).asJava
+  private def asAVMap[K](kvs: (Symbol, K)*)(implicit fk: DynamoFormat[K]) =
+    Map(kvs: _*).map { case (k,v) => (k.name, fk.write(v)) }.asJava
```
adamnfish: You use the `->` sugar elsewhere, why not here as well?

src/main/scala/com/gu/scanamo/ScanamoRequest.scala
```diff
@@ -41,22 +41,22 @@ object ScanamoRequest {
...
-  def queryRequest[K](tableName: String)(key: (String, K))(implicit fk: DynamoFormat[K]): QueryRequest = {
+  def queryRequest[K](tableName: String)(key: (Symbol, K))(implicit fk: DynamoFormat[K]): QueryRequest = {
     val (k, v) = key
     new QueryRequest().withTableName(tableName)
-      .withKeyConditionExpression(s"#K = :$k")
-      .withExpressionAttributeNames(Map("#K" -> k).asJava)
-      .withExpressionAttributeValues(asAVMap(s":$k" -> v))
+      .withKeyConditionExpression(s"#K = :${k.name}")
+      .withExpressionAttributeNames(Map("#K" -> k.name).asJava)
+      .withExpressionAttributeValues(asAVMap(Symbol(s":${k.name}") -> v))
```
adamnfish: Shame there's no `mapKeys` method.

adamnfish:
Much better!

:+1:

src/main/scala/com/gu/scanamo/ScanamoRequest.scala
philwills: When I'm unpacking and repacking a tuple, I like the symmetry of having the bracket form on both sides.  If you could have the `->` sugar in the pattern match, then I'd do that.

LATaylor-guardian:
@philwills I think it makes perfect sense here as a representation for an identifier. Although it does imply some knowledge of what a `Symbol` is so you lose a little clarity somewhat for those not familiar. 

:+1:

src/main/scala/com/gu/scanamo/ScanamoRequest.scala
LATaylor-guardian: I'm assuming implementing filtering on a query request `withFilterExpression` would follow a similar pattern?

philwills: I'm basically rewriting all of this to support more interesting queries at the moment, but yes, they should hopefully have a similar interface.

PR #5, opened 2016-03-14
Separate get and delete methods for different kinds of key

### LINKED ISSUE #4: Support different types for hash key and range key
`get`, `delete` and so on take the key as `(key: (Symbol, K)*)`, but this means that all the components of the primary key have to have the same type `K`. This seems like quite a special case to me.

For example, in my app, the hash key is a `RecipeId` (with a DynamoFormat to persist it as a string) and the range key is the build number, an `Int`.

What about just providing 2 versions of each method that requires a key? e.g.

```scala
def get[K, T](client: AmazonDynamoDB)(tableName: String)(hashkey: (Symbol, K))
    (implicit fk: DynamoFormat[K], ft: DynamoFormat[T]): Option[ValidatedNel[DynamoReadError, T]] = ...

def get[HK, RK, T](client: AmazonDynamoDB)(tableName: String)(hashkey: (Symbol, HK), rangekey: (Symbol, RK))
    (implicit fhk: DynamoFormat[HK], frk: DynamoFormat[RK], ft: DynamoFormat[T]): Option[ValidatedNel[DynamoReadError, T]] = ...
```

The varargs param strikes me as a bit odd anyway, as you will only ever pass exactly 1 or 2 keys.

I can put a PR together if you're happy with this suggestion.

cb372:
One `get` method that takes only a hash key, another that takes both a hash key and a range key. Same for `delete`.

Fixes #4

Poke @philwills

### DIFF
--- a/src/main/scala/com/gu/scanamo/Scanamo.scala
+++ b/src/main/scala/com/gu/scanamo/Scanamo.scala
@@ -12,7 +12,8 @@ import com.gu.scanamo.DynamoResultStream.{QueryResultStream, ScanResultStream}
   * {{{
   * >>> import com.amazonaws.services.dynamodbv2.model._
   * >>> val client = LocalDynamoDB.client()
-  * >>> val createTableResult = LocalDynamoDB.createTable(client, "farmers", List("name" -> ScalarAttributeType.S), List("name" -> KeyType.HASH))
+  * >>> val createFarmersTableResult = LocalDynamoDB.createTable(client, "farmers", List("name" -> ScalarAttributeType.S), List("name" -> KeyType.HASH))
+  * >>> val createFarmHandsTableResult = LocalDynamoDB.createTable(client, "farmhands", List("boss" -> ScalarAttributeType.S, "employeeId" -> ScalarAttributeType.N), List("boss" -> KeyType.HASH, "employeeId" -> KeyType.RANGE))
   * }}}
   */
 object Scanamo {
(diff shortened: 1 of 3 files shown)


src/main/scala/com/gu/scanamo/Scanamo.scala
```diff
@@ -62,9 +81,28 @@ object Scanamo {
     * None
     * }}}
     */
-  def delete[K, T](client: AmazonDynamoDB)(tableName: String)(key: (Symbol, K)*)
-    (implicit fk: DynamoFormat[K], ft: DynamoFormat[T]): DeleteItemResult =
-    client.deleteItem(deleteRequest(tableName)(key: _*))
+  def delete[HK, T](client: AmazonDynamoDB)(tableName: String)(hashkey: (Symbol, HK))
+    (implicit fhk: DynamoFormat[HK], ft: DynamoFormat[T]): DeleteItemResult =
```
cb372: @philwills The `T` type param and `ft` implicit param are not necessary, could be removed?

cb372:
The only problem is that overloading the methods like this confuses the compiler, meaning you need to specify type parameters for it to work out which overloaded method you are calling. This really shouldn't be necessary.

```
scala> val farmer: Option[ValidatedNel[DynamoReadError, Farmer]] = Scanamo.get(client)("farmers")('name -> "Fred")
<console>:18: error: ambiguous reference to overloaded definition,
both method get in object Scanamo of type [HK, RK, T](client: com.amazonaws.services.dynamodbv2.AmazonDynamoDB)(tableName: String)(hashkey: (Symbol, HK), rangekey: (Symbol, RK))(implicit fhk: com.gu.scanamo.DynamoFormat[HK], implicit frk: com.gu.scanamo.DynamoFormat[RK], implicit ft: com.gu.scanamo.DynamoFormat[T])Option[cats.data.ValidatedNel[com.gu.scanamo.DynamoReadError,T]]
and  method get in object Scanamo of type [HK, T](client: com.amazonaws.services.dynamodbv2.AmazonDynamoDB)(tableName: String)(hashkey: (Symbol, HK))(implicit fhk: com.gu.scanamo.DynamoFormat[HK], implicit ft: com.gu.scanamo.DynamoFormat[T])Option[cats.data.ValidatedNel[com.gu.scanamo.DynamoReadError,T]]
match argument types (com.amazonaws.services.dynamodbv2.AmazonDynamoDB)
       val farmer: Option[ValidatedNel[DynamoReadError, Farmer]] = Scanamo.get(client)("farmers")('name -> "Fred")
```

(`Scanamo.get[String, Farmer](client)("farmers")('name -> "Fred")` compiles just fine.)

I can't see a way round this other than giving the methods different names, but I couldn't think of any good ones. Open to suggestions.

philwills:
So you might be interested in https://github.com/guardian/scanamo/compare/query-conditions?expand=1. I started building some machinery for doing queries and then realised the same machinery could be used for eliminating the type parameters from the `get` operation.

I'm slightly worried that I've gone too far down the implicit rabbit-hole, but I think I can at least remove some of the cruddy inheritance.

philwills:
My further adventures down the rabbit-hole, led to https://github.com/guardian/scanamo/compare/query-condition-typeclass?expand=1

cb372:
@philwills That is beautiful :+1: :shipit:

PR #6, opened 2016-03-20
Add support for putting multiple items at at time

philwills:


src/main/scala/com/gu/scanamo/Scanamo.scala
```diff
@@ -36,6 +36,23 @@ object Scanamo {
...
+    * >>> case class Rabbit(name: String)
+    * >>> import com.amazonaws.services.dynamodbv2.model.ScalarAttributeType._
+    * >>> val createTableResult = LocalDynamoDB.createTable(client)("rabbits")('name -> S)
+    * >>> val multiPut = Scanamo.putAll(client)("rabbits")((
+    * ...   for { _ <- 0 until 100 } yield Rabbit(util.Random.nextString(500))).toList).toList
+    * >>> Scanamo.scan[Rabbit](client)("rabbits").toList.size
+    * 100
+    * }}}
+    */
+  def putAll[T](client: AmazonDynamoDB)(tableName: String)(items: List[T])(implicit f: DynamoFormat[T]): Streaming[BatchWriteItemResult] =
```
adamnfish: I realise it's a bit late for this feedback but:

* why are there so many argument lists? It's a bit harder to use in IntelliJ this way because you can't just `Ctrl+P` to see what the method needs.
* Why isn't the client implicit? It seems like needless verbosity to have to provide it every time

Much simpler:
```scala
def putAll[T](tableName: String, items: List[T])(implicit client: AmazonDynamoDB, f: DynamoFormat[T])
```

Further thought suggests maybe you want to do currying for the client and table. Since Scala's support for this is terrible and you aren't using that approach in the docstrings, I don't feel like you're ever going to see people using it that way. That would be an excellent idea in Haskell.

philwills: My idea was that at some later point I might put a layer over the top, by exploiting currying the function. I haven't fleshed out how it would work and as you say, support can be a bit iffy, so that may turn out to be a dead end.

I'm currently sceptical of using implicit arguments for things like client here. Especially when mixing them with typeclasses. I don't think the verbosity hurts that much and I'd rather try other approaches to get round it.

P.S. Thanks, more than happy to have feedback late. There's no release with this in yet and even if there were, the API is definitely not stable enough for me to get too stressed about changing it if something doesn't look like working out.

PR #8, opened 2016-03-29
Support multiple key get, delete, query, etc.

philwills:
Also supports queries which are not just exact matches, e.g. `<` and `beginsWith` and gets rid of some unnecessary type parameters.

I like the consistency of the interface, but I'm less keen on the amount of implicit machinery involved.

### DIFF
--- a/src/main/scala/com/gu/scanamo/Scanamo.scala
+++ b/src/main/scala/com/gu/scanamo/Scanamo.scala
@@ -107,9 +107,9 @@ object Scanamo {
     * None
     * }}}
     */
-  def delete[K, T](client: AmazonDynamoDB)(tableName: String)(key: (Symbol, K)*)
+  def delete[K, T](client: AmazonDynamoDB)(tableName: String)(key: (Symbol, K))
     (implicit fk: DynamoFormat[K], ft: DynamoFormat[T]): DeleteItemResult =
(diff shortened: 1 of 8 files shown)


src/main/scala/com/gu/scanamo/Scanamo.scala
```diff
@@ -80,21 +100,38 @@ object Scanamo {
...
+    * {{{
+    * >>> import com.gu.scanamo.syntax._
+    * >>> import com.amazonaws.services.dynamodbv2.model.ScalarAttributeType._
+    *
+    * >>> case class Doctor(actor: String, regeneration: Int)
+    * >>> val doctorsTableResult = LocalDynamoDB.createTable(client)("doctors")('actor -> S, 'regeneration -> N)
+    *
+    * >>> val putDoctorResult = Scanamo.putAll(client)("doctors")(
+    * ...   List(Doctor("McCoy", 9), Doctor("Ecclestone", 10), Doctor("Ecclestone", 11)))
+    * >>> Scanamo.getAll[Doctor](client)("doctors")(('actor and 'regeneration) -> List("McCoy" -> 9, "Ecclestone" -> 11))
```
philwills: I don't really like this syntax, but it was the least bad option I could think of.

I'd love to hear of better alternatives.

PR #9, opened 2016-03-31
Doctest the README

philwills:
Also remove tut, as for now I think the README is more useful than a site.


cb372:
:+1: makes sense.

I didn't know about sbt-doctest before, although I know the guy who wrote it :)

PR #14, opened 2016-04-15
Provide a simplified means of creating a DynamoFormat

philwills:
The common case is that there is some potential exception thrown when converting back from what was stored in Dynamo


philwills:
This was the pattern for all the custom formats used in https://github.com/guardian/riff-raff/blob/8a03aeddc514d4183057aa4d015fa7d4e12d5577/riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala#L16

PR #20, opened 2016-04-26
remove discipline dependency

### LINKED ISSUE #18: Unused discipline dependency
Scanamo depends on discipline but doesn't seem to use it. It's also not in test scope. Is it safe to remove, if only for smaller jar size?

joprice:
Fixes #18


philwills:
Thanks!

PR #21, opened 2016-04-26
Add a Table type

philwills:
This aim is for this to be the main way of using Scanamo. That does rely on people being comfortable with using the `Free` type.

cc @adamnfish

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,11 +103,13 @@ scala> def temptWithGum(child: LuckyWinner): LuckyWinner = child match {
      | }
 scala> val luckyWinners = Table[LuckyWinner]("winners")
 scala> val operations = for {
-     |      _           <- luckyWinners.putAll(
-     |                       List(LuckyWinner("Violet", "human"), LuckyWinner("Augustus", "human"), LuckyWinner("Charlie", "human")))
-     |      winners     <- luckyWinners.scan()
-     |      _           <- luckyWinners.putAll(winners.flatMap(_.toOption).map(temptWithGum).toList)
-     |      results     <- luckyWinners.getAll('name -> List("Charlie", "Violet"))
+     |      _               <- luckyWinners.putAll(
+     |                           List(LuckyWinner("Violet", "human"), LuckyWinner("Augustus", "human"), LuckyWinner("Charlie", "human")))
+     |      winners         <- luckyWinners.scan()
+     |      winnerList      =  winners.flatMap(_.toOption).toList
+     |      temptedWinners  =  winnerList.map(temptWithGum)
+     |      _               <- luckyWinners.putAll(temptedWinners)
+     |      results         <- luckyWinners.getAll('name -> List("Charlie", "Violet"))
      | } yield results
      
 scala> Scanamo.exec(client)(operations).toList
--- a/README.md
+++ b/README.md
@@ -94,19 +94,24 @@ scala> import com.gu.scanamo.syntax._
 
 scala> val client = LocalDynamoDB.client()
 scala> import com.amazonaws.services.dynamodbv2.model.ScalarAttributeType._
-scala> val farmersTableResult = LocalDynamoDB.createTable(client)("free")('name -> S)
+scala> val farmersTableResult = LocalDynamoDB.createTable(client)("winners")('name -> S)
 
-scala> case class Free(name: String, number: Int)
+scala> case class LuckyWinner(name: String, shape: String)
+scala> def temptWithGum(child: LuckyWinner): LuckyWinner = child match {
+     |   case LuckyWinner("Violet", _) => LuckyWinner("Violet", "blueberry")
+     |   case winner => winner
+     | }
+scala> val luckyWinners = Table[LuckyWinner]("winners")
 scala> val operations = for {
-     |      _           <- ScanamoFree.putAll("free")(List(Free("Monad", 1), Free("Applicative", 2), Free("Love", 3)))
-     |      maybeMonad  <- ScanamoFree.get[Free]("free")('name -> "Monad")
-     |      monad       = maybeMonad.flatMap(_.toOption).getOrElse(Free("oops", 9))
-     |      _           <- ScanamoFree.put("free")(monad.copy(number = monad.number * 10))
-     |      results     <- ScanamoFree.getAll[Free]("free")('name -> List("Monad", "Applicative"))
+     |      _           <- luckyWinners.putAll(
+     |                       List(LuckyWinner("Violet", "human"), LuckyWinner("Augustus", "human"), LuckyWinner("Charlie", "human")))
+     |      winners     <- luckyWinners.scan()
+     |      _           <- luckyWinners.putAll(winners.flatMap(_.toOption).map(temptWithGum).toList)
+     |      results     <- luckyWinners.getAll('name -> List("Charlie", "Violet"))
      | } yield results
      
 scala> Scanamo.exec(client)(operations).toList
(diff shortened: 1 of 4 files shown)


joprice:
I was just telling someone that I was going to contribute this. Thanks for reading my mind!

src/main/scala/com/gu/scanamo/Table.scala
```diff
@@ -0,0 +1,126 @@
...
+  def put(v: V) = ScanamoFree.put(name)(v)
+  def putAll(vs: List[V]) = ScanamoFree.putAll(name)(vs)
+  def get(key: UniqueKey[_]) = ScanamoFree.get[V](name)(key)
+  def getAll(keys: UniqueKeys[_]) = ScanamoFree.getAll[V](name)(keys)
+  def delete(key: UniqueKey[_]) = ScanamoFree.delete(name)(key)
+}
+
+private[scanamo] case class Index[V: DynamoFormat](tableName: String, indexName: String)
+
+/* typeclass */trait Scannable[T[_], V] {
```
adamnfish: Is there any significance to this very strange formatting?

src/main/scala/com/gu/scanamo/package.scala
```diff
@@ -4,7 +4,7 @@ import com.gu.scanamo.query._
 
 package object scanamo {
 
-  object syntax {
+  object syntax extends Scannable.ToScannableOps with Queryable.ToQueryableOps {
```
adamnfish: Why use traits and inheritance here instead of importing the operations from objects?

src/main/scala/com/gu/scanamo/Table.scala
```diff
@@ -0,0 +1,126 @@
...
+object Scannable {
+  def apply[T[_], V](implicit s: Scannable[T, V]) = s
+
+  trait Ops[T[_], V] {
+    val instance: Scannable[T, V]
+    def self: T[V]
+    def scan() = instance.scan(self)()
+  }
+
+  trait ToScannableOps {
```
adamnfish: As below, why are these traits rather than objects?

README.md
```diff
@@ -94,19 +94,24 @@ scala> import com.gu.scanamo.syntax._
...
+     |   case winner => winner
+     | }
+scala> val luckyWinners = Table[LuckyWinner]("winners")
 scala> val operations = for {
-     |      _           <- ScanamoFree.putAll("free")(List(Free("Monad", 1), Free("Applicative", 2), Free("Love", 3)))
-     |      maybeMonad  <- ScanamoFree.get[Free]("free")('name -> "Monad")
-     |      monad       = maybeMonad.flatMap(_.toOption).getOrElse(Free("oops", 9))
-     |      _           <- ScanamoFree.put("free")(monad.copy(number = monad.number * 10))
-     |      results     <- ScanamoFree.getAll[Free]("free")('name -> List("Monad", "Applicative"))
+     |      _           <- luckyWinners.putAll(
```
adamnfish: Better :-)

README.md
```diff
@@ -94,19 +94,24 @@ scala> import com.gu.scanamo.syntax._
...
+     |                       List(LuckyWinner("Violet", "human"), LuckyWinner("Augustus", "human"), LuckyWinner("Charlie", "human")))
+     |      winners     <- luckyWinners.scan()
+     |      _           <- luckyWinners.putAll(winners.flatMap(_.toOption).map(temptWithGum).toList)
```
adamnfish: Hiding this (important) `map` at the end of a long line is a bit confusing. Maybe do the map on its own line above? It's the key piece of functionality in the example so it should be clear to someone glancing at it that it's one of the "steps" (alongside the scan, put and getAll).

build.sbt
```diff
@@ -27,6 +27,8 @@ scalacOptions := Seq(
   "-feature",
   "-unchecked",
   "-language:implicitConversions",
+  "-language:higherKinds",
+  "-language:existentials",
```
adamnfish: https://www.youtube.com/watch?v=siwpn14IE7E :-D

src/main/scala/com/gu/scanamo/Table.scala
```diff
@@ -0,0 +1,126 @@
...
+  def index(indexName: String) = Index[V](name, indexName)
+
+  def put(v: V) = ScanamoFree.put(name)(v)
+  def putAll(vs: List[V]) = ScanamoFree.putAll(name)(vs)
+  def get(key: UniqueKey[_]) = ScanamoFree.get[V](name)(key)
+  def getAll(keys: UniqueKeys[_]) = ScanamoFree.getAll[V](name)(keys)
+  def delete(key: UniqueKey[_]) = ScanamoFree.delete(name)(key)
+}
+
+private[scanamo] case class Index[V: DynamoFormat](tableName: String, indexName: String)
```
adamnfish: I don't understand why this is private given that there are public facing methods that essentially take an `Index` (e.g. `withTableWithSecondaryIndex`)?

adamnfish:
Great change, I've a few comments on it but mainly looks really good.

The completely unnecessary multi-parameter list style is still filling me with rage though ;-)

src/main/scala/com/gu/scanamo/Table.scala
philwills: It was reminding me of the fact that these are supposed to behave as if I was using the `@typeclass` annotation from simulacrum, but it is a bit weird, so I'll probably remove it.

src/main/scala/com/gu/scanamo/package.scala
philwills: Just importing an op from another object doesn't export into the implicit scope when importing `object._`

README.md
philwills: Agreed. I plan to do some more work on the README, but hopefully this is a bit clearer now.

src/main/scala/com/gu/scanamo/Table.scala
philwills: `withTableWithSecondaryIndex` isn't part of the public API, that's only used in the tests which aren't published. Making `Index` private gives me a bit more implementation wiggle room without risking breaking things, but I don't feel enormously strongly about it.

src/main/scala/com/gu/scanamo/package.scala
adamnfish: Really?! It works fine with e.g. implicit formats?

README.md
```diff
@@ -94,19 +94,26 @@ scala> import com.gu.scanamo.syntax._
...
-     |      _           <- ScanamoFree.putAll("free")(List(Free("Monad", 1), Free("Applicative", 2), Free("Love", 3)))
-     |      maybeMonad  <- ScanamoFree.get[Free]("free")('name -> "Monad")
-     |      monad       = maybeMonad.flatMap(_.toOption).getOrElse(Free("oops", 9))
-     |      _           <- ScanamoFree.put("free")(monad.copy(number = monad.number * 10))
-     |      results     <- ScanamoFree.getAll[Free]("free")('name -> List("Monad", "Applicative"))
+     |      _               <- luckyWinners.putAll(
+     |                           List(LuckyWinner("Violet", "human"), LuckyWinner("Augustus", "human"), LuckyWinner("Charlie", "human")))
+     |      winners         <- luckyWinners.scan()
+     |      winnerList      =  winners.flatMap(_.toOption).toList
+     |      temptedWinners  =  winnerList.map(temptWithGum)
```
adamnfish: +100

src/main/scala/com/gu/scanamo/Table.scala
adamnfish: Ah ok, I was thinking of the docstring tests as the public-facing instructions but you're just using them for all the tests. This is reasonable then!

PR #24, opened 2016-04-28
Add implicit DynamoFormat for Double and Set

williamho:
This adds some implicit formats for
* `Set[Int]`, `Set[Long]`, and `Set[Double]` (using `getNS`/`withNS`) 
* `Set[String]` (using `getSS`/`withSS`)

Not sure if there's a better way to make the `Set` implicits be more generic, since sets only exist for numbers and strings (rather than `AttributeValue`s)


philwills:
Thanks @williamho, this looks great.

philwills:
I've just pushed a 0.4.0 release with these changes included.

williamho:
Thanks! :tada:

PR #31, opened 2016-05-11
Support conditional `put`

philwills:


src/main/scala/com/gu/scanamo/request/ScanamoPutRequest.scala
```diff
@@ -0,0 +1,15 @@
+package com.gu.scanamo.request
+
+import com.amazonaws.services.dynamodbv2.model.AttributeValue
+
+case class ScanamoPutRequest(
```
philwills: I was a bit loathe to move away from using the underlying request objects directly, but their lack of a copy constructor makes working with them really painful for the type of transformations needed for compound conditionals.

src/main/scala/com/gu/scanamo/Table.scala
```diff
@@ -61,6 +61,103 @@ case class Table[V: DynamoFormat](name: String) {
...
+    * List(Right(Thing(b,Some(3))), Right(Thing(c,Some(42))), Right(Thing(a,None)))
+    *
+    * >>> case class Compound(a: String, maybe: Option[Int])
+    * >>> val compoundTable = Table[Compound]("compounds")
+    * >>> LocalDynamoDB.withTable(client)("compounds")('a -> S) {
+    * ...   val ops = for {
+    * ...     _ <- compoundTable.putAll(List(Compound("alpha", None), Compound("beta", Some(1)), Compound("gamma", None)))
+    * ...     _ <- compoundTable.given(attributeExists('maybe) and 'a -> "alpha").put(Compound("alpha", Some(2)))
+    * ...     _ <- compoundTable.given(attributeExists('maybe) and 'a -> "beta").put(Compound("beta", Some(3)))
+    * ...     _ <- compoundTable.given(Condition('a -> "gamma") and attributeExists('maybe)).put(Compound("gamma", Some(42)))
```
philwills: Having to use `Condition()` if it's the first part of a compound condition is ugly. I'm hoping to find a better alternative, but I this is just about good enough to put out there.

Issue #28, opened 2016-05-09
How to limit results size when using query


kraus-as24:
Is there a way to limit the result size using `ScanamoAsync.query`?

Thanks in advance!

philwills:
It returns a stream, which lazily pulls in page by page, so if you for instance `.take(100)`  on it, then it will only pull back the number of pages necessary to support that. 

It will always attempt to pull back a full page of results, so you might pull back a bit more data than is strictly necessary, but there shouldn't be additional requests.

kraus-as24:
Thanks. I actually need only the first result (or the last one using a descending query) with ordering given by a range key. But as DynamoDB is supporting defining a limit on queries is it planned to add this feature? And what do you think is the difference concerning performance. Thanks.

philwills:
It's not the near the top of my list at the moment, as my main focus with the library is to try and allow things to be done in a simple, safe way and this is currently achievable, if not optimal. Any suggestions or PRs on how it could be supported without major changes to the API would be welcome.

In terms of performance, it's very difficult to know in the abstract. I'd be a bit surprised if it was a bottleneck, but without benchmarking the particular app I wouldn't like to pretend certainty.

adamnfish:
One concern is that in an app with a single-read-heavy workload, you'd have to over-provision your tables by `pagesize`. I wonder if exposing a way of setting the page size on the underlying DynamoDB req would give the developer the chance to optimise requests for the expected use-case?

I'll have a look through the code, see if I can work out a nice way to do it.

philwills:
That's a good point about capacity consumption. It's now definitely on the list.

philwills:
This is now available in 0.5.0

PR #34, opened 2016-05-13
Support limiting the results coming back from a query

philwills:
Addresses #28

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,6 +55,8 @@ doctestDecodeHtmlEntities := true
 doctestWithDependencies := false
 doctestTestFramework := DoctestTestFramework.ScalaTest
 
+parallelExecution in Test := false
+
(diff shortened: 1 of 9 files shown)


philwills:
During implementation of this, I realised that the `Stream`s being returned were eagerly evaluated.

I haven't found a good way to avoid that yet, so for the moment this changes everything to return `List` instead.

cb372:
I also had a quick look at getting Stream to play nicely with Free but didn't have any luck

philwills:
This https://github.com/typelevel/cats/blob/34e206af4b061730587095399ee2d3f0e8e0ed97/core/src/main/scala/cats/std/stream.scala#L45 was pointed out to me and there was a suggestion I might want to look at CoFree. I decided at that point that, for the moment at least, discretion would be the better part of valour.

PR #36, opened 2016-05-17
Add format for Vector

### LINKED ISSUE #26: Pattern for supporting other collection types
I attempted to define a `DynamoFormat` for scala's `Vector`, using listFormat as a template and found that the some useful helper functions (`attribute`, `javaListFormat`) are marked private. It would be convenient have access to these and avoid duplicating them in client code, or have an alternative helper for users that want to customize collection serialization.

philwills:
Also made javaListFormat public as it was requested and I don't see a good reason for keeping it private.

Fixes #26


joprice:
👏

PR #48, opened 2016-06-24
Removed `Requests` object

philwills:
Plus a few other suggested tweaks form @janua and @adamnfish


src/main/scala/com/gu/scanamo/ops/ScanamoInterpreters.scala
```diff
@@ -0,0 +1,130 @@
...
+import com.amazonaws.AmazonWebServiceRequest
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.services.dynamodbv2.{AmazonDynamoDB, AmazonDynamoDBAsync}
+import com.amazonaws.services.dynamodbv2.model.{UpdateItemResult, _}
+import com.gu.scanamo.request.{ScanamoDeleteRequest, ScanamoPutRequest, ScanamoUpdateRequest}
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success}
+
+object ScanamoInterpreters {
```
adamnfish: Why not just `Interpreters`? The package tells us that it is a Scanamo thing. (I need an emoji for "this is a minor point")

src/main/scala/com/gu/scanamo/ScanamoFree.scala
```diff
@@ -1,48 +1,56 @@
...
+          new WriteRequest().withPutRequest(new PutRequest().withItem(f.write(i).getM))
+        ).asJava).asJava)
+      )
+    )
 
   def get[T](tableName: String)(key: UniqueKey[_])
     (implicit ft: DynamoFormat[T]): ScanamoOps[Option[Xor[DynamoReadError, T]]] =
     for {
-      res <- ScanamoOps.get(getRequest(tableName)(key))
+      res <- ScanamoOps.get(new GetItemRequest().withTableName(tableName).withKey(key.asAVMap.asJava))
```
janua: So just to be clear; this `new` is because we don't have `GetItemRequest` in `scanamo`... yet. Might be worth eventually adding all the verbs/requests for consistency?

philwills: I was originally trying to avoid reinventing them, but working with mutable objects when building up things like conditions was painful. I think you're right that ideally it would be more consistent.

src/main/scala/com/gu/scanamo/ops/ScanamoInterpreters.scala
philwills: Mainly because I'm addicted to IntelliJ's 'find class', which is much less fun to use if I have to worry about the package as well as the name.
 
Am willing to be convinced otherwise, but that's my rationale for preferring the duplication/verbosity.

Issue #50, opened 2016-07-14
Batch GET requests to scanamo fail with duplicate keys


lindseydew:
Currently if a 'getAll' method is called is a non distinct list this is the error we see:
AWS Error Code: ValidationException, AWS Error Message: Provided list of item keys contains duplicates

philwills:
I'm counting this as more or less fixed by #51

PR #51, opened 2016-07-18
Use Sets for bulk operations

philwills:
As noted in #50, DynamoDB assumes that bulk operations are performed on a Set. This doesn't ensure the set is of the form expected by DynamoDB, i.e. it doesn't ensure the hash/range key combination are unique, but I think it better conveys the intent.

cc @lindseydew @emma-p


emma-p:
👍

lindseydew:
Nice one 👍

PR #52, opened 2016-07-20
add format for Seq via List

paulmr:


philwills:
Thanks @paulmr, would you mind adding a test for this? I've been using [sbt-doctest](https://github.com/tkawachi/sbt-doctest#usage) primarily and it's property based tests in particular for the `DynamoFormat`s which I think has worked well.

Let me know if you need a hand with this.

paulmr:
apologies for the delay @philwills , I have now added a test for this.

philwills:
Brilliant, thanks!

PR #54, opened 2016-08-03
Update to Cats 0.7.0

travisbrown:
This updates the library to the Cats 0.7.0 snapshot, and shouldn't be merged until 0.7.0 is published. It also depends on [this Cats change](https://github.com/typelevel/cats/pull/1261), which is merged but not yet in the snapshot, so you'll have to publish the snapshot locally if you want to run the tests right now (they do pass if you do that).

I've also changed the cats dependency from `cats` to `cats-free`, since the `cats` module is [only meant to be a pedagogical convenience](https://github.com/typelevel/cats/issues/1181#issuecomment-231589592) (it pulls in test dependencies, etc.), and shouldn't be used in production.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,19 +5,21 @@ scalaVersion := "2.11.8"
 
 crossScalaVersions := Seq(scalaVersion.value, "2.10.6")
 
+resolvers += Resolver.sonatypeRepo("snapshots")
+
 libraryDependencies ++= Seq(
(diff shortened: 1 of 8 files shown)


philwills:
Lovely, thanks. 

I think when I originally pulled in Cats, it was when everything had been pulled back into a single module, but I'm happy to just depend on cats-free/cats-core now that they're split apart again.

travisbrown:
I've rebased and updated the PR for the 0.7.0 release—thanks again, @philwills.

PR #57, opened 2016-08-10
adding byteArrayFormat and byteFormat backed by AttributeValues' Byte…

drocsid:
…Buffer

https://github.com/guardian/scanamo/issues/55

https://github.com/guardian/scanamo-scrooge/issues/4

### DIFF
--- a/src/main/scala/com/gu/scanamo/DynamoFormat.scala
+++ b/src/main/scala/com/gu/scanamo/DynamoFormat.scala
@@ -177,17 +177,16 @@ object DynamoFormat extends DerivedDynamoFormat {
     */
   implicit val shortFormat = xmap(coerceNumber(_.toShort))(_.toString)(numFormat)
 
-  // Thrift and therefore Scanamo-Scrooge provides a byte and binary types backed by byte and byte[]
-  // Since AttributeValue includes a ByteBuffer instance, creating byte formats backed by ByteBuffer
+  // Thrift and therefore Scanamo-Scrooge provides a byte and binary types backed by byte and byte[].
+  implicit val byteFormat = xmap(coerceNumber(_.toByte))(_.toString)(numFormat)
 
+  // Since AttributeValue includes a ByteBuffer instance, creating byteArray format backed by ByteBuffer
   private val javaByteBufferFormat = attribute[java.nio.ByteBuffer](_.getB, "B")(_.withB)
 
   private def coerceByteBuffer[B](f: ByteBuffer => B): ByteBuffer => Xor[DynamoReadError, B] =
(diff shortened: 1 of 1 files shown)


src/main/scala/com/gu/scanamo/DynamoFormat.scala
```diff
@@ -176,6 +177,17 @@ object DynamoFormat extends DerivedDynamoFormat {
...
+  // Since AttributeValue includes a ByteBuffer instance, creating byte formats backed by ByteBuffer
+
+  private val javaByteBufferFormat = attribute[java.nio.ByteBuffer](_.getB, "B")(_.withB)
+
+  private def coerceByteBuffer[B](f: ByteBuffer => B): ByteBuffer => Xor[DynamoReadError, B] =
+    coerce[ByteBuffer,B, IllegalArgumentException](f)
+
+  implicit val byteArrayFormat = xmap(coerceByteBuffer(_.array()))(a => ByteBuffer.wrap(a))(javaByteBufferFormat)
+
+  implicit val byteFormat = xmap(coerceByteBuffer(_.get(0)))(b => ByteBuffer.wrap(Array(b)))(javaByteBufferFormat)
```
philwills: I'd be tempted to store individual bytes as a number, in the same way we do with Short, rather than in a ByteBuffer of one byte. Do you have a particular preference either way?

philwills:
Thanks for this, I've raised a question about how to store an individual `Byte`, but generally this looks good.

drocsid:
Your suggestion makes sense. I wasn't sure about the cost of a ByteBuffer, but after reading http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html , I think that a Number might be the least expensive way to store the [bB]yte. I was being lazy and worried about conversions.

drocsid:
I'll try to update the PR tomorrow, unless you want to implement your suggestions. It's 4:14 AM here and I don't have my home laptop ready to test.

philwills:
I might take a look, but definitely won't be staying up till 4:14.

philwills:
This look great, thanks @drocsid

drocsid:
@philwills thank you for this project. What's the easiest way that I can pull these into my project backed by scanamo-scrooge, so I can test against https://github.com/guardian/scanamo-scrooge/issues/4 ? Do I pull .6.1-SNAPSHOT in from a local build? Sorry for my lack of knowledge.

drocsid:
Figured out it's sbt publishLocal . Just learning to use sbt, and it wasn't in the getting started guide. Well hopefully I can close out the issues and move forward.

PR #60, opened 2016-09-18
Add a map method to DynamoFormat

cb372:
This is a purely selfish PR. Given:

```
case class RoleId(value: String) extends AnyVal
```

it means I can write:

```
implicit val dynamoFormat: DynamoFormat[RoleId] =
  DynamoFormat.map[RoleId, String](RoleId.apply)(_.value)
```

instead of:

```
implicit val dynamoFormat: DynamoFormat[RoleId] =
  DynamoFormat.xmap[RoleId, String](value => Xor.right(RoleId(value)))(_.value)
```

### DIFF
--- a/src/main/scala/com/gu/scanamo/DynamoFormat.scala
+++ b/src/main/scala/com/gu/scanamo/DynamoFormat.scala
@@ -52,7 +52,7 @@ import java.nio.ByteBuffer
   * Right(LargelyOptional(None,Some(X)))
   * }}}
   *
-  * Custom formats can often be most easily defined using [[DynamoFormat.coercedXmap]] or [[DynamoFormat.xmap]]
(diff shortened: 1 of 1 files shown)


philwills:
It feels a bit odd to call it `map` when it goes both ways, but supporting the case where the domains match exactly seems sensible.

cb372:
I can't think of a good name for a `map` that goes both ways. `iso`? `transform`?

philwills:
I think either are fine. I'd probably err towards `iso` as that distinguishes it from the other cases nicely, but am slightly worried it might be too obscure.

LATaylor-guardian:
`reversibleMap`?

LATaylor-guardian:
`biMap`?

philwills:
I quite like `biMap` except that it might get confused with the one from https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/functor/Bifunctor.scala#L12  or its ilk, which seems to be the more common use of that name.

PR #63, opened 2016-09-19
Add REMOVE support to the update API

cb372:
Note that this is different from the already implemented `DELETE`.

* `DELETE` is for deleting a given element from a Set attribute
* `REMOVE` is for removing an attribute from an item


philwills:
Brilliant, thanks @cb372 

I do feel a bit bad for Harry though.

PR #64, opened 2016-09-20
Provide better errors when can't derive a DynamoFormat

philwills:
Add a universal macro at the bottom of the priority tree that reports the specific missing DynamoFormat, not just the top-level case class.

This is a REPL session showing the behaviour:

image

Previously you would just get that last line.

### DIFF
--- a/src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
+++ b/src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
@@ -3,12 +3,13 @@ package com.gu.scanamo
 import cats.data.{NonEmptyList, Validated, Xor}
 import com.amazonaws.services.dynamodbv2.model.AttributeValue
 import com.gu.scanamo.error.{DynamoReadError, InvalidPropertiesError, MissingProperty, PropertyReadError}
+import macrocompat.bundle
 import shapeless._
 import shapeless.labelled._
 
 import collection.convert.decorateAll._
 
-trait DerivedDynamoFormat {
+trait DerivedDynamoFormat extends NoFormatError {
   type ValidatedPropertiesError[T] = Validated[InvalidPropertiesError, T]
 
   trait ConstructedDynamoFormat[T] {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
```diff
@@ -55,3 +56,23 @@ trait DerivedDynamoFormat {
       def write(t: T): AttributeValue = formatR.value.write(gen.to(t))
     }
 }
+
+trait NoFormatError {
+  import scala.language.experimental.macros
+  implicit def moreHelpfulErrorMessage[T]: DynamoFormat[T] = macro MacroOfLastResort.moreHelpfulErrorMessage[T]
+}
+
+@bundle
```
LATaylor-guardian: What does @bundle do?

cb372:
Nice! 👍

src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
```diff
@@ -55,3 +56,23 @@ trait DerivedDynamoFormat {
...
+class MacroOfLastResort(val c: scala.reflect.macros.blackbox.Context) {
```
LATaylor-guardian: What does `scala.reflect.macros.blackbox.Context` represent?

src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
philwills: See http://docs.scala-lang.org/overviews/macros/bundles.html

Also get used by https://github.com/milessabin/macro-compat if you want cross-compilation.

src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
philwills: My understanding is that it's essentially the state of compilation. So for instance it can give you the `enclosingPosition` which is the point of code from which the macro was called and can be used for things like aborting  macro.

It's also responsible for determining the compilation universe you're in, which I'm not going to pretend I understand well enough to give a coherent explanation of.

LATaylor-guardian: That's great, thanks.

cb372: It can also be thought of as your entry point to the reflection API. Importing `c.universe._` brings into scope everything your macro will need, including:

* quasiquotes (`q"..."`)
* all the AST types such as `TermName`, `ClassDef`, `Apply`, `Select`
* methods for looking up/manipulating types, e.g. `weakTypeOf`, `appliedType`, `lub`
* Useful debugging methods such as `showRaw` and `showCode`

PR #65, opened 2016-09-20
Read DynamoDB NULL as None

philwills:
This is complicated by the fact that the Java client's API returns a boxed Boolean which is `null` if no NULL value is set. Using an explicit `Boolean.unbox`, so that `null` evaluates to `false`is the least clumsy workaround I can find for this.


philwills:
`x => x.isNULL == null || !x.isNULL` is a possible alternative, which might be a little more explicit.

philwills:
Addresses https://github.com/guardian/scanamo/issues/49

PR #67, opened 2016-10-25
`Xor` -> `Either`

philwills:
With `Either` becoming left-biased in Scala 2.12, cats is moving away from `Xor`. This is a very literal replacement to try and stay ahead of that change.

As part of a future change, I'd like to represent more errors as values, but this change doesn't address that.


rtyley:
> With Either becoming left-biased in Scala 2.12

That's right-biased, right?

https://github.com/typelevel/cats/issues/1192
http://www.scala-lang.org/api/2.12.x/scala/util/Either.html

philwills:
@rtyley yes, you're right, it's right

rtyley:
Right

PR #69, opened 2016-11-09
2.12 support

philwills:


mchv:
Nice! not so many changes at the end

philwills:
@mchv there are a few more things that can probably be changed to improve things, like moving from `traverseU` to `traverse`, but I thought I'd keep this minimal to get support out.

Issue #27, opened 2016-05-03
Support batch deletion


joprice:
There is `putAll` and `getAll`, would `deleteAll` make sense as well? This doc also demonstrates a mix of puts and deletes:

http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/batch-operation-document-api-java.html

Perhaps there could be syntax for mixing the two like `batch(put = List(...), delete = List(...))`

philwills:
A `deleteAll` would be a good addition. 

Mixing the two seems a bit odd to me and I haven't seen any compelling uses in the wild, so I'm not yet convinced it's worth the complexity.

joprice:
I have at one use case for it so far that corresponds to a user on a mobile device modifying a list of items and sending a batch of modifications and removals. But it's probably a rare use case.

randallalexander:
@philwills we have a working version of deleteAll based off the putAll that we would be willing to contribute back.  My main question is what do you think the result should be?  Our version tries to go through the result and figure out what was actually deleted and return that.  (Since the AWS result does return unprocessed items https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html)  Alternatively we could just return the unprocessed items, the BatchWriteItemResult, or any other variant you can think of.  I am open to whatever you think is more inline with the library design.  Let me know and I will work on a PR for it.

philwills:
This is now supported as of 0.8.2

PR #70, opened 2016-11-21
batch deleteAll

randallalexander:
Proposal for batch deleteAll
  - resolves item #27

### DIFF
--- a/src/main/scala/com/gu/scanamo/ScanamoFree.scala
+++ b/src/main/scala/com/gu/scanamo/ScanamoFree.scala
@@ -4,7 +4,7 @@ import com.amazonaws.services.dynamodbv2.model.{PutRequest, WriteRequest, _}
 import com.gu.scanamo.DynamoResultStream.{QueryResultStream, ScanResultStream}
 import com.gu.scanamo.error.DynamoReadError
 import com.gu.scanamo.ops.ScanamoOps
-import com.gu.scanamo.query._
+import com.gu.scanamo.query.{UniqueKey, _}
 import com.gu.scanamo.request.{ScanamoDeleteRequest, ScanamoPutRequest, ScanamoUpdateRequest}
 import com.gu.scanamo.update.UpdateExpression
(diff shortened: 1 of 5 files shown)


philwills:
Thanks for this.  I'll try to have a proper look later, but first impressions look good.

src/main/scala/com/gu/scanamo/ScanamoFree.scala
```diff
@@ -14,18 +14,34 @@ object ScanamoFree {
...
-    items.grouped(25).toList.traverseU(batch =>
+    items.grouped(batchSize).toList.traverseU(batch =>
       ScanamoOps.batchWrite(
         new BatchWriteItemRequest().withRequestItems(Map(tableName -> batch.toList.map(i =>
           new WriteRequest().withPutRequest(new PutRequest().withItem(f.write(i).getM))
         ).asJava).asJava)
       )
     )
 
+  def deleteAll[T](tableName: String)(deleteItems: (Symbol, Set[T]))(implicit f: DynamoFormat[T]): ScanamoOps[List[BatchWriteItemResult]] = {
```
philwills: I think it would be a bit more consistent for this to take a `Set[UniqueKey[_]]`.

randallalexander: I was trying `Set[UniqueKey[_]]` and ran across `UniqueKeys[_]` which I think is a better fit.  about to push an update.  Let me know what you think.

randallalexander:
I believe I have addressed everything.  Not sure if you wanted me to update the main README.md too.  Assuming "no" but you know what they say about assumptions.

philwills:
Wonderful, thanks @randallalexander!

philwills:
This is now released in `0.8.2`, which should close #27.

I'm currently working on overhauling the documentation, so haven't added anything to the README yet, but didn't think it was necessary to tie that to the release.

PR #71, opened 2016-11-30
Adding consistentRead flag option for Get operations

amherrington13:


amherrington13:
Related issue here: [https://github.com/guardian/scanamo/issues/72](url)

philwills:
Thanks @amherrington13, this definitely seems like a useful feature.

In general, I've tried to err away from additional parameters on these methods, so I'd like to have a think if there's an alternative way to construct this which would achieve the same goals. Perhaps something analogous to the way scans/queries can be limited https://github.com/guardian/scanamo/blob/master/src/main/scala/com/gu/scanamo/Table.scala#L198 might work, but I don't have any definite idea.

Also, I'm pleased to hear that you've successfully been using it.

philwills:
Hmm, looking at this, I wonder whether your original approach might have been better after all. How do you feel about how the two approaches have panned out?

Thanks again for putting in the time to do this.

amherrington13:
Hmm, I do like that this approach stays consistent with the way limit is being done for scans/queries. It would also set us up nicely to add the same strong consistency option to the Scannable and Queryable objects, as well as make further extension to Get operations easier in the future. But the first approach did feel a bit more simplistic. I really could go either way. What are your thoughts?

I'm glad to be able to contribute. Our team is really enjoying using scanamo.

mikquinlan:
+1.

We're using this library at Ticketmaster and have the need for a consistent read operation.

gsson:
When providing configuration overrides, since many of default overrides in a specific context will be reused it would be nice not to have to repeat it.

One option would be to provide the `get` consistent read option (and other future options) as implicits with default values.

Silly examples with one configuration and one typeclass based approach
```
import org.scalatest.{FlatSpec, Matchers}

class ScanamoTest extends FlatSpec with Matchers {

  // Analog to com.amazonaws.services.dynamodbv2.model.GetItemRequest
  case class ReadOperation(consistentReads: Boolean = false) {
    def apply(): String = {
      if (consistentReads) {
        "strongly consistent reads"
      }
      else {
        "eventually consistent reads"
      }
    }
  }

  // Config approach

  case class ReadConfiguration(consistentReads: Boolean)

  object DefaultReadConfiguration$ extends ReadConfiguration(false)

  // Analog to Scanamo*.get
  def configRead()(implicit config: ReadConfiguration = DefaultReadConfiguration$): String = {
    ReadOperation(config.consistentReads)()
  }

  it should "use the provided config" in {
    configRead() should be ("eventually consistent reads")
  }

  it should "use the provided config" in {
    implicit val config = ReadConfiguration(consistentReads = true)
    configRead() should be ("strongly consistent reads")
  }

  // Type class approach

  trait OperationModifier[T] {
    def apply(t: T): T
  }

  object OperationModifier {
    def nullModifier[T]: OperationModifier[T] = new OperationModifier[T] {
      override def apply(t: T): T = t
    }
  }

  object WithConsistentReads extends OperationModifier[ReadOperation] {
    override def apply(t: ReadOperation): ReadOperation = {
      t.copy(consistentReads = true) // Like GetItemRequest.withConsistentRead(true)
    }
  }

  // Analog to Scanamo*.get
  def modifierRead()(implicit operationModifier: OperationModifier[ReadOperation] = OperationModifier.nullModifier): String = {
    operationModifier(ReadOperation())()
  }


  it should "use the null operation modifier" in {
    modifierRead() should be ("eventually consistent reads")
  }

  it should "use the provided operation modifier" in {
    implicit val operationModifier = WithConsistentReads
    modifierRead() should be ("strongly consistent reads")
  }
}
```

gsson:
I added a consistent read variant based on the type class example above; https://github.com/guardian/scanamo/compare/master...gsson:ConsistentRead

I'm thinking i might be overcomplicating things :)

(and in the end, as long as I get consistent reads I'm good -- I just got inspired by @philwills comment)

amherrington13:
Hmm, looking at this. I like the idea of a RequestModifier typeclass. To my point about adding the consistency option for scans and queries in the future (as well as any other request modification options), this would make that a trivial thing. Between the options we have, I don’t have too strong a preference either way. I, too, will be good as long as we get consistent reads.

philwills:
I'm going to merge this now and have opened #74, which adds Scan/Query variants.

In the future I'd like to gradually expose less of the AWS API directly, so I'm less keen on the RequestModifier approach. 

Thanks for all your input.

gsson:
Thanks!

amherrington13:
Thanks!

philwills:
Support for consistent Get/Scan/Query is out as 0.8.3. Please let me know how you get on.

PR #81, opened 2017-01-05
Add default format for UUID

philwills:
This seems a very common use case, at least at the Guardian, and this seems a reasonable default.


cb372:
👍 This is a line that I often end up writing as well

PR #82, opened 2017-01-09
update readme to latest release version

gidsg:


cb372:
You might want to add a release step to automate this. See https://github.com/cb372/scalacache/blob/master/build.sbt#L189-L199 for an example.

PR #85, opened 2017-01-10
Delete 'more helpful' errors, as they can be confusing

philwills:
See https://github.com/guardian/scanamo/issues/76


travisbrown:
I'm reluctantly :+1: on this.

PR #90, opened 2017-01-24
Fix bug when many updates and'd together

philwills:
Previously the repeated search and replace of a string could lead to multiple parameters having the same placeholder in the expression string.


philwills:
I'm sure there could be a less repetitive implementation, but this works and is mentally simpler for me at least.

PR #94, opened 2017-02-21
The put action returns the item that was put in the table

susiecoleman:
Previously a `PutItemResult` was returned when an item was put in a table. Now the item that has been inserted is returned


philwills:
Closing this after in person conversation with @susiecoleman

PR #95, opened 2017-02-22
[WIP] Expose request query

ayoub-benali:
An attempt to expose RequestQuery in scanamo. The goal is to be able to add filters when querying
(with the java API) which is not yet possible with the scanamo DSL


ayoub-benali:
sorry it was a mistake, the PR was meant for my fork ..

Issue #104, opened 2017-04-25
Implement 'Between' RangeKeyCondition


todor-kolev:
Similar to [`BeginsWith`](https://github.com/guardian/scanamo/blob/9cb47068bc3f69709c5bdb1a792dd9a6d5562e01/src/main/scala/com/gu/scanamo/query/DynamoKeyCondition.scala#L42) it would be great to have a `Between` RangeKeyCondition.

https://github.com/guardian/scanamo/blob/9cb47068bc3f69709c5bdb1a792dd9a6d5562e01/src/main/scala/com/gu/scanamo/query/DynamoKeyCondition.scala#L26

http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Condition.html

todor-kolev:
https://github.com/guardian/scanamo/pull/106

PR #106, opened 2017-05-02
#104 implement 'between'

todor-kolev:
* Added `Between` `RangeKeyCondition`
* Added `betweenCondition` `ConditionExpression`
* Added some DSL so you can do stuff like `'mode -> "Underground" and ('zone between (2 and 4))`


philwills:
This looks excellent, thanks @todor-kolev

Issue #103, opened 2017-04-11
getAll requires manual batching


philwills:
Currently `putAll` operations work around the maximum batch size limitation of DynamoDB by batching the operations behind the scenes, but inconsistently `getAll` requires the user to do the batching themselves.

`getAll` should manage the batching so that the user of the library doesn't have to worry about it.

todor-kolev:
https://github.com/guardian/scanamo/pull/116

philwills:
This is now out in `0.9.5`

PR #116, opened 2017-06-02
#103 Automatically managing batching for getAll

todor-kolev:
Although it can be improved, this fixes the DynamoDB batch get size limitation of 100 items per request.

### DIFF
--- a/src/main/scala/com/gu/scanamo/ScanamoFree.scala
+++ b/src/main/scala/com/gu/scanamo/ScanamoFree.scala
@@ -55,16 +55,14 @@ object ScanamoFree {
     } yield
       Option(res.getItem).map(read[T])
 
-
   def getAll[T: DynamoFormat](tableName: String)(keys: UniqueKeys[_]): ScanamoOps[Set[Either[DynamoReadError, T]]] = {
-    for {
-      res <- ScanamoOps.batchGet(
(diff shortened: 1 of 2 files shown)


src/main/scala/com/gu/scanamo/ScanamoFree.scala
```diff
@@ -55,16 +55,14 @@ object ScanamoFree {
...
+    keys.asAVMap.grouped(batchSize).toList.traverseU { batch =>
+      ScanamoOps.batchGet(
         new BatchGetItemRequest().withRequestItems(Map(tableName ->
-          new KeysAndAttributes().withKeys(keys.asAVMap.map(_.asJava).asJava)
+          new KeysAndAttributes().withKeys(batch.map(_.asJava).asJava)
         ).asJava)
       )
-    } yield
-      res.getResponses.get(tableName).asScala.toSet.map(read[T])
+    }.map(_.flatMap(_.getResponses.get(tableName).asScala.toSet.map(read[T])).toSet)
```
philwills: This line is now doing so much that I struggle with reading it, but I'm happy to merge for now and revisit later.

todor-kolev: I agree, I will try to refactor it when I get a chance.

PR #117, opened 2017-06-14
Add ability to do equality check on secondary keys in query dsl.

keshav-elevio:
Let me know


keshav-elevio:
build failed waiting for dynamo to start on travis

philwills:
Sorry, only just noticed this. `ConditionExpression` which you've changed here, isn't used for querying, that's `QueryableKeyCondition`. In particular, the condition on the range key is specified with a `RangeKeyCondition`. It's `RangeKeyCondition` that would need to be added to.

Also, the general pattern used for specifying a key equals something has to be to use a `(Symbol, V)` tuple, rather than having an explicit `equals` method.

regiskuckaertz:
Hi @keshav-elevio, feel free to raise a PR for `QueryableKeyCondition`.

PR #125, opened 2017-06-28
Minimal config scalafmt

philwills:
This aims to make the formatting of Scanamo more consistent by using scalafmt. 

At present it still requires a manual formatting step.

The config is fairly minimal at the moment with the idea that subsequent pull requests can make incremental changes which can be seen on their own merits.


philwills:
I don't love all of these changes, but I don't think any of them are heinous enough to sacrifice consistency for.

LATaylor-guardian:
Is ScalaFmt worth using by default from now on in projects over Scalariform? I thought Scalariform 'just worked'. What are the benefits of ScalaFmt over Scalariform?

PR #149, opened 2017-10-04
Add `returnValues` support for put/update/delete operations

amirkarimi:
By setting `ReturnValues` for put, update and delete requests we will be able to get the item attributes as they appeared before they were updated.

### DIFF
--- a/src/main/scala/com/gu/scanamo/ScanamoAsync.scala
+++ b/src/main/scala/com/gu/scanamo/ScanamoAsync.scala
@@ -1,7 +1,7 @@
 package com.gu.scanamo
 
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsync
-import com.amazonaws.services.dynamodbv2.model.{BatchWriteItemResult, DeleteItemResult, PutItemResult}
+import com.amazonaws.services.dynamodbv2.model.{BatchWriteItemResult, DeleteItemResult, PutItemResult, ReturnValue}
 import com.gu.scanamo.error.DynamoReadError
 import com.gu.scanamo.ops.{ScanamoInterpreters, ScanamoOps}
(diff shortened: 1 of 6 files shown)


philwills:
Thanks for this @amirkarimi, a few thoughts:

 * I'd rather not add anything to `Scanamo` or `ScanamoAsync`, see https://github.com/guardian/scanamo/pull/146
 * I've tried to avoid default arguments in Scanamo, `limit` (https://github.com/guardian/scanamo/blob/master/scanamo/src/main/scala/com/gu/scanamo/Table.scala#L271) on `Table` is an example of an alternative approach
 * `update` already has similar functionality, but returns the fully hydrated object, rather than the `Map[String,AttributeValue]`, see http://guardian.github.io/scanamo/latest/api/com/gu/scanamo/Table.html#update(key:com.gu.scanamo.query.UniqueKey[_],expression:com.gu.scanamo.update.UpdateExpression):com.gu.scanamo.ops.ScanamoOps[Either[com.gu.scanamo.error.DynamoReadError,V]]. I'd rather follow that pattern of trying to hide DynamoDB client internals.

amirkarimi:
@philwills My problem with `update` is that I can't just pass an object to it to be upserted (though it's possible to create an UpdateExpression using DynamoFormat?). Anyway, do you suggest making `put` like `update` to return the fully hydrated object?

philwills:
Yes, I was suggesting that returning the fully hydrated object from `put` would be an alternative approach that I believe would fitter with the wider design. 

I wasn't suggesting that `update` was a suitable alternative for this use case.

amirkarimi:
@philwills Thanks for the explanation. I'll resolve the conflicts and follow this approach.

amirkarimi:
As we decided to implement this feature in another way I just deleted the previous branch and created a new one from `master`. Deletion caused this PR to be closed. Anyway, this PR is continued at #153

PR #151, opened 2017-10-05
Split project into multi-modules

btlines:
Split scanamo into. multi-modules:

* formats: the dynamo formats (conversion to/from `AttributeValue`)
* scanamo: the core module (same as before without the formats)
* alpakka: A new interpreter based on Alpakka (c.f. https://github.com/guardian/scanamo/issues/145)


build.sbt
```diff
@@ -69,24 +104,47 @@ lazy val scanamo = (project in file("scanamo"))
...
+      "org.scalacheck" %% "scalacheck" % "1.13.4" % Test
+    ),
+
+    fork in Test := true,
+    envVars in Test := Map(
+      "AWS_ACCESS_KEY_ID" -> "dummy",
+      "AWS_SECRET_KEY" -> "credentials"
+    ),
+    dynamoDBLocalDownloadDir := file(".alpakka-dynamodb-local"),
+    dynamoDBLocalPort := 8052,
```
btlines: Wish I don't need this but when it can conflicts with `scanamo` module tests when both modules tests are run at the same time.
There's probably a way to run the tests sequentially between modules though

philwills:
Thanks @btlines, this is great.

PR #158, opened 2017-10-21
Update documentation for the asynchronous interpreters

calvinlfer:
Make the `docs` subproject depend on `scanamo` and `alpakka` subproject as we need Akka's `ActorSystem` and `ActorMaterializer` as they are dependencies needed for the Alpakka client to function.

**Note:** Don't bring in test files from alpakka as they interfere with scanamo's test files causing the Local DynamoDB to point to the wrong port.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -156,7 +156,7 @@ lazy val docs = (project in file("docs"))
     addMappingsToSiteDir(mappings in (ScalaUnidoc, packageDoc) in root, siteSubdirName in ScalaUnidoc in root)
   )
   .enablePlugins(MicrositesPlugin, SiteScaladocPlugin, GhpagesPlugin)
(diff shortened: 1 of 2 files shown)


calvinlfer:
Closes https://github.com/guardian/scanamo/issues/145

calvinlfer:
I also found that `scanamo/alpakka/src/main/scala/com/gu/scanamo/ScanamoAlpakka.scala` has an error that prevents a publish from going through. I was trying to do local testing with `publishLocal` and I found that Scaladocs weren't able to be generated because of `[[com.gu.scanamo.Scanamo]]`. It appears that you can't reference the files in the core `scanamo` project.

philwills:
Thanks @calvinlfer. This looks great and I appreciate the heads up about the publishing issue.

Issue #136, opened 2017-09-04
Support for nested fields in conditional expressions


ivashin:
Currently Scanamo supports conditions only on top-level item fields. It would be great to be able to use nested fields in conditional expressions similar to nested field updates.

I've reused Field definition from nested updates and put up a small proof of concept for the simplest equality testing use-case:

```scala
object NestedFieldTest {
  case class Record(id: String, nested: Nested)
  case class Nested(s: String, i: Int)
}

class NestedFieldTest extends FunSpec with Matchers {
  import NestedFieldTest._

  lazy val client = AmazonDynamoDBClientBuilder.standard().withEndpointConfiguration(new EndpointConfiguration("http://localhost:8000", "")).build()
  lazy val table = Table[Record]("TestTable")

  implicit def fieldValueEqualsCondition[V: DynamoFormat] = new ConditionExpression[(Field, V)] {
    override def apply(pair: (Field, V))(condition: Option[RequestCondition]): RequestCondition =
      RequestCondition(
        s"#${pair._1.placeholder} = :condition",
        pair._1.attributeNames.map { case (k, v) => s"#$k" -> v },
        Some(Map(s":condition" -> DynamoFormat[V].write(pair._2)))
      )
  }

  describe("Scanamo") {
    val id = "ABC"
    val putRequest = table.put(Record(id, Nested("A", 0)))
    Scanamo.exec(client)(putRequest)

    it("should allow to use nested field in condition expressions") {
      val updateRequest = table
        .given('nested \ 's -> "A")
        .update('id -> id, set('nested \ 's -> "B") and set('nested \ 'i -> 42))
      val updated = Scanamo.exec(client)(updateRequest)
      updated.right.value should equal(Record(id, Nested("B", 42)))
    }

    it("should combine condition expressions with nested fields") {
      val updateRequest = table
        .given(Condition('nested \ 's -> "B") and ('nested \ 'i -> 42))
        .update('id -> id, set('nested \ 'i -> 1000))
      val updated = Scanamo.exec(client)(updateRequest)
      updated.right.value should equal(Record(id, Nested("B", 1000)))
    }
  }
}
```

This will probably fail for more complex scenarios, such as nested name conflicts (e.g. `'a \ 'c` and `'b \ 'c` in one combined expression), but should be not hard to fix and extend to other conditions.

philwills:
This is out in `1.0.0-M2`

ivashin:
[...]

PR #165, opened 2017-10-30
Equals condition support for nested attributes

ivashin:
Fixes remaining issue in #136. Slight addition to #156, which adds an implicit conversion from `(AttributeName, V)` to `ConditionExpression`, necessary to support nested fields. Not the prettiest code, but I could not find a solution to replace `Symbol, V)` conversion completely.


philwills:
Looks good to me.

Might you be able to add a test case for the new behaviour? Completely understand if you don't have time, but since I'm a bit stretched at the moment, it might help move things along.

ivashin:
Done. Should have do it in the first place but didn't have time then.

philwills:
Great, thanks @ivashin

PR #173, opened 2017-12-05
Make space for back-off algorithm

regiskuckaertz:
Thought I would try to add support for incomplete writes. Ultimately one will want to decide exactly how to handle unprocessed items at the interpreter level, as the behaviour will depend on the destination monad. 

The cleanest way I can think of is by introducing a new operation in the algebra (`BatchWriteRetry`) that can be treated separately. From there, it is not hard to add back-off with jitter and maximum retries.

### DIFF
--- a/alpakka/src/main/scala/com/gu/scanamo/ops/AlpakkaInterpreter.scala
+++ b/alpakka/src/main/scala/com/gu/scanamo/ops/AlpakkaInterpreter.scala
@@ -22,6 +22,7 @@ object AlpakkaInterpreter {
           case Query(req) => client.single(JavaRequests.query(req))
           case Update(req) => client.single(JavaRequests.update(req))
           case BatchWrite(req) => client.single(req)
+          case BatchWriteRetry(req) => client.single(req)
           case BatchGet(req) => client.single(req)
           case ConditionalDelete(req) =>
             client.single(JavaRequests.delete(req))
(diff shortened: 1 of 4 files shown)


regiskuckaertz:
The build broke for some unrelated error 🤔 

I wonder if the signature of `putAll` should reflect the fact that each batch may fail, i.e. return a `(BatchWriteResult, Option[BatchWriteResult])` (or a `Cofree[List, BatchWriteResult]`)

philwills:
Sorry for being so slow to respond. 

My instinct is that the retrying logic should be at the interpreter level, rather than in the Free grammar. I tend to think that making the grammar simpler and the interpreters richer make for the more flexible API.

regiskuckaertz:
No worries. Yes, I agree, the idea of having a spurious constructor did not sit well with me anyway. I thought about that initially, the way I saw it there were two solutions:

- use monadic composition, i.e. `Free.FlatMapped`, to recursively execute `batchWrite` if there are unprocessed items. That is the solution I wrote here.

- write a cofree interpreter. Intuitively, the interpreter would need the ability to inject new instructions in the program based on the result of the current instruction. However, [`foldMap`](https://github.com/typelevel/cats/blob/master/free/src/main/scala/cats/free/Free.scala#L152-L157) is not powerful enough; in fact, this process of expanding an execution tree is unfold instead of a fold.

I'll give that one a try 😃

regiskuckaertz:
Closing for now.

PR #179, opened 2018-01-04
Upgrade 0.9.x to cats 1.0.1

tabdulradi:
Hey, my team is till on 0.9.x, but the old Cats dependency conflicts with other libs, so I cherry-picked couple of commits to bring the latest Cats to 0.9 folks.

Can you please create a branch from `v0.9.5` tag so that I can target my PR against?


tabdulradi:
Also should I bump the version `0.9.6` as part of this PR or do I manage versioning separately?

regiskuckaertz:
You can grab M7 which depends on Cats v1.1.0

PR #182, opened 2018-01-12
Add a module to support refined types

alonsodomin:
Hi all, hope you find this interesting.

At work we use refined types in some projects so we end up writing wiring (integration?) with other libraries that do not support it. Most of the time the code is as trivial as in this case, so I took the liberty of adding a module to the project to support them so everyone else can also profit from it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -86,6 +86,20 @@ lazy val formats = (project in file("formats"))
     doctestTestFramework := DoctestTestFramework.ScalaTest
   )
 
+lazy val refined = (project in file("refined"))
+  .settings(
+    commonSettings,
+    publishingSettings,
+    name := "scanamo-refined"
+  )
+  .settings(
+    libraryDependencies ++= Seq(
+      "eu.timepit" %% "refined" % "0.8.6",
+      "org.scalatest" %% "scalatest" % "3.0.4" % Test
+    )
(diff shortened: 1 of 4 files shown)


alonsodomin:
it seems that there is no much interest in this piece of code... I wouldn't mind to release it under a different `groupId` but it would be great to get some sort of feedback...

philwills:
Sorry, not lack of interest, just haven't been finding time to keep up. Thankfully some of my former colleagues are planning to help out.

philwills:
Thanks @alonsodomin, this looks great.

alonsodomin:
@philwills wow, that was fast! thanks for considering! :)

philwills:
@davidfurey did  a release, but I failed to notice that this wasn't aggregated into the root SBT project, so it hasn't been released. I think it should be possible to release just this module by checking out the tag and publishing from within the refined project, but it may be safer just to do another release.

I've opened #194 to hopefully mean it will consistently get released in future.

davidfurey:
@philwills - I tried just publishing the refined module but it didn't work.  If there is no objections I'll just do another release rather than spend time investigating why it failed.

philwills:
:+1:

davidfurey:
Done

PR #183, opened 2018-01-15
Handle the case in which the attribute value is not a map in derived format

miguel-vila:
Currently this would fail with a null pointer exception:

```scala
case class Thing(foo: String, bar: Int)
DynamoFormat[Thing].read(new AttributeValue().withS("whatevs")) // throws NPE

sealed trait Base
case class Foo(f1: String) extends Base
case class Bar(f2: Int) extends Base
DynamoFormat[Base].read(new AttributeValue().withS("whatevs")) // throws NPE
```

This PR should fix it and return an error.

I can add tests if required.

### DIFF
--- a/formats/src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
+++ b/formats/src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala
@@ -91,13 +91,25 @@ trait DerivedDynamoFormat {
 
   implicit def genericProduct[T: NotSymbol, R](implicit gen: LabelledGeneric.Aux[T, R], formatR: Lazy[ValidConstructedDynamoFormat[R]]): DynamoFormat[T] =
     new DynamoFormat[T] {
-      def read(av: AttributeValue): Either[DynamoReadError, T] = formatR.value.read(av).map(gen.from).toEither
+      def read(av: AttributeValue): Either[DynamoReadError, T] =
+        if(av.getM == null) {
+          Left(TypeCoercionError(new Exception(s"Expected $av to be a map")))
+        } else {
+          formatR.value.read(av).map(gen.from).toEither
+        }
+
       def write(t: T): AttributeValue = formatR.value.write(gen.to(t))
     }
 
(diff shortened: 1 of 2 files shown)


philwills:
Sorry have been rubbish at keeping up with things lately. This looks sensible. A test would be lovely, but not required.

miguel-vila:
Added two simple tests which should reproduce the problem. Also modified the way it's checked that the `AttributeValue` is a map. I had initially done this inside the derived implementation but the problem with doing this is that in the hcons case we are restricted to return an `InvalidPropertiesError` which may not make sense to do here, given that it's not just a specific property error, but a structure error. Let me know what you think.

johnduffell:
I've had a quick look at this, it's a good spot.  I was wondering if it would be more appropriate to move the checking to the point where it's used rather than doing it preemptively?

More specifically, I believe that we could do the null check just before the asScala conversions, to keep the java->scala boundary in one place. i.e. `Option(av.getM).map(_.asScala)`  Then we can decide what to do in that one place.

It would be on lines:
https://github.com/scanamo/scanamo/blob/10faff7c922e039e2f80b5426239c365181a44a5/formats/src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala#L39
https://github.com/scanamo/scanamo/blob/10faff7c922e039e2f80b5426239c365181a44a5/formats/src/main/scala/com/gu/scanamo/DerivedDynamoFormat.scala#L74

Any thoughts about this?

miguel-vila:
@johnduffell that was my initial approach. Problem is that for the hcons case we are restricted to return errors of type `InvalidPropertiesError` which I'm not sure is the adequate one for this situation.

I think we can start by deciding which error we want to return. We probably want this error to be the same in both cases. My idea was a single `TypeCoercionError(new Exception(s"Expected $av to be a map"))` error. Any opinion about this?

regiskuckaertz:
Hello @miguel-vila , I'm open to widen the error channel in `ValidatedPropertiesError[T]` and `ConstructedDynamoFormat[T]` to `DynamoReadError`, so that this limitation is lifted and you can move check at the calling site. If you have appetite to do that, we can get this merged! Thanks a lot for your contribution.

regiskuckaertz:
100% onboard with that change, so feel free to reopen when/if you have appetite for it.

PR #184, opened 2018-01-16
Implement get all with consistent reads

Slakah:
Implements [consistent reads](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html) for the `getAll` operation (`getAllWithConistency`).

Note: the tests appear to only test that the `WithConsistency` methods function correctly, and don't directly test that the `KeysAndAttributes.consistentRead` field is set. This could be tested using a test interpreter which would capture the resultant `BatchGetItemRequest` and test that `consistentRead` is set.


philwills:
Sorry for being so slow to respond. This looks great, thanks!

philwills:
This should now be available in 1.0.0-M5

PR #185, opened 2018-01-24
Write None as an av with null set to true

Slakah:
Currently `DynamoFormat[Option[T]]` will write a `None` as an `(AttributeValue) null`, instead of `new AttributeValue().withNULL(true)`. I suspect that this bug only really affects the Dynamo DAX clients, but we have seen errors when using a `DynamoFormat` which produces a field with a `None`.

## TODO

- [x] Write a test to prove that dynamo accepts an av of the form:
```
new AttributeValue().withM(Map("foo" -> new AttributeValue().withNULL(true))
```
I suspect it doesn't, which means [mapFormat](https://github.com/scanamo/scanamo/blob/master/formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala#L391) will need to be updated.


philwills:
I think I did originally encode None as  `new AttributeValue().withNULL(true)`, but ran into some problem which meant I switched to just using null. Unfortunately, I'm struggling to dredge up what that problem was.

regiskuckaertz:
Can we get this one out along with #250? They to tackle similar issues.I think it just needs a couple of tests.

Slakah:
@regiskuckaertz, sorry for the late response (been away on holiday). I've added an addition test https://github.com/scanamo/scanamo/pull/185/commits/ee82473f15d7c09208e5dad9d1b00cb3ef30af5c, it copies the work done for https://github.com/scanamo/scanamo/blob/1055ccda4f91b9808c72fd12ce7ae9566a753615/formats/src/test/scala/com/gu/scanamo/DynamoFormatTest.scala .

Might need some rejigging of the PRs to get them merged.

It appears as though the travis test failure is unrelated, but I'm unable to retrigger.

regiskuckaertz:
That's weird. I re-ran it and now we get an error message: an empty string cannot be inserted into Dynamo. I suppose the `Gen[Option[String]]` should filter out `Some("")` to solve the issue.

regiskuckaertz:
Hi @Slakah , can you rebase and fix the failing test so we can get this one out? Thanks for all your work!

PR #187, opened 2018-01-26
Allow append/prepend of list values

prurph:
Adds two new functions: `appendList` and `prependList`. Each takes
a list argument and appends or prepends it to an existing value at
the specified attribute name.

The previous methods, `append` and `prepend`, are unchanged: they
take a value, cast it to a list of a single element, and then
append/prepend it. This does not allow for list concatenation;
appending a list would result in a nested list, thus likely causing
any custom deserialization that expected a flat list for the updated
attribute to fail.

I messed around with ways to allow the existing `append` to take either a list or a single argument with varargs and things like that but couldn't get anything working. Since the argument is a tuple I wasn't able to do something clever like overload `def append[V: DynamoFormat](fieldValue: (AttributeName, V*)` or things like that.

### DIFF
--- a/scanamo/src/main/scala/com/gu/scanamo/Table.scala
+++ b/scanamo/src/main/scala/com/gu/scanamo/Table.scala
@@ -170,6 +170,25 @@ case class Table[V: DynamoFormat](name: String) {
     * List(Right(Character(James Bond,List(Craig))))
     * }}}
     *
+    * To concatenate a list to the front or end of an existing list, use appendAll/prependAll:
+    *
+    * {{{
+    * >>> case class Fruit(kind: String, sources: List[String])
+    * >>> val fruits = Table[Fruit]("fruits")
+    *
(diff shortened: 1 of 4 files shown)


philwills:
Sorry for the delay responding. This looks good. Would `appendAll/prependAll` be a bit more conventional, or is that just in my head?

alexflav23:
Hi @philwills Any chance getting a release for this in the near future? Can do a local in the mean time, but I don't think there's currently a way to append to a collection.

prurph:
@philwills `appendAll`, `prependAll` sounds better to me too. I've gone ahead and changed it.

philwills:
Great, thanks @prurph 

Yes, I'll try and work with @davidfurey to get a release out before too long, but I can't promise a date.

philwills:
This should now be available in 1.0.0-M5

prurph:
Great thank you!

PR #190, opened 2018-02-22
Fix broken documentation links

howardjohn:
Many links to the scaladocs were linking to a url with a /scanamo
prefix. However, the main documentation is just hosted at /, causing
broken links.

Ran `grep -r "/scanamo"` to ensure I picked up all the links and built the docs to make sure the links worked now.


philwills:
Brilliant, thanks!

PR #193, opened 2018-03-15
Add instructions for releasing

philwills:


philwills:
Yea, I think that would be worthwhile.

davidfurey:
Done, although the documentation update failed because I don't have jekyll installed.  I'll try to deploy the docs only on Monday.

PR #208, opened 2018-04-25
Add ScanamoAsync + Cats Effect Interop

emilypi:
This PR adds functionality for `Effect` types from `cats.effect` as a base functionality in `ScanamoAsync`, including the relevant interpreters. I've kept and moved the `Future`-based async functionality to a new package `async`, which allows the individual to use a construct called `FutureAsync` if they require `Future`s so as not to put anyone out. I've also provided the relevant tests for the new `Effect`-based `ScanamoAsync`.

Note: I would like to consider deprecating `Future` usage, as this code should be considered legacy in light of more modern options. This is for both performance reasons, as well as code quality and pollution concerns (think `ExecutionContexts` polluting every api as a result of this design choice). We can provide an `Effect` instance for `Future` if need be, in the same way Monix has produced an `Effect` instance for `monix.eval.Task`, and in the same vein, we have for Scalaz's new Bifunctor-based `IO` monad. This change would give maximum interoperability between Scanamo and Scala's robust 3rd party ecosystem, and promote more responsible coding styles.

Also, y'all needed a version bump on `cats`  😄


emilypi:
@philwills Hi Phil - I tried pinging in the Gitter to discuss this, but I didn't get a rsponse, so I'm messaging here with the raised PR.

philwills:
Sorry for the lack of response, will try and take a look

emilypi:
Thanks @philwills!

philwills:
Thanks for the really well put together PR. I definitely agree that it would be good to support making interaction with DynamoDB a managed effect.

Having said that I'm not comfortable with making a change as radical as this before hitting a 1.0 release, which I realise has been slow coming. As far as I know cats-effect doesn't have any API stability promises yet. 

What I would be delighted with now is having this as an opt in module like Scanamo-alpakka. If you would like to attempt that, it would be very welcome, but if not, I'll try to use this as a basis for doing it, though I can't make any promises on when.

emilypi:
Thanks for the kind words @philwills - I'd be very willing to support an additional module for both `cats-effect` and `scalaz-ioeffect` (not that we've backported our new IO work). I'll close this PR and migrate the work to a separate one that provides the intended modules.

emilypi:
Closing this in favor of https://github.com/scanamo/scanamo/pull/212

Issue #105, opened 2017-04-25
'Table#getAll' should handle empty 'UniqueKeys' set gracefully


todor-kolev:
Currently `Table#getAll` would result in `com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException: The list of keys in RequestItems for BatchGetItem is required: my_table has empty list (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: ...)` if the keys set is empty. 

It would be better if it silently returns an empty list so you don't have to perform a check before calling `Table#getAll`

regiskuckaertz:
Hi @todor-kolev 

First of all, our apologies for not handling this issue until now; the news agenda has kept us busy to say the least 😅

I’ve given this issue some thought. I think you are right that scanamo should provide some measure of protection against a subset of errors. However I wouldn’t want our solution to encourage bad practices; besides, an empty list might not be semantically correct depending on if you view the bottle as half-full or half-empty.

A nice solution IMHO would be to leverage the type system so that this constraint can be statically checked at compile time. Do you feel like this would be an acceptable fix for your use case?

todor-kolev:
I don't think there's a way to check how many key ids the client will ask for at compile time but I guess you can make the function accept NonEmptyList  of some sort. .

PR #213, opened 2018-05-13
Use NonEmptySet for key conditions

regiskuckaertz:
Addresses #105 

Cats 1.1.0 provides a `NonEmptySet` abstraction that can be used to avoid `BatchGetItem` queries that would error in the database. Because these are _ordered_ sets, instances for `Order[V]` and `Order[(H, R)]` must be available in scope, meaning the odd cats-kernel import will be required, e.g.
```scala
import cats.data.NonEmptySet
import cats.kernel.instances.string._

NonEmptySet.of("Boggis", "Bean")
```

---

I've thought about another solution to this problem, as I felt the `Order` constraint was completely unnecessary here. The only thing we need is a way to reify a non-empty `Set` similar to `NonEmptySet.of(h: A, t: A*)`. It turns out marrying this idea with tuples, which are central to the Scanamo DSL, is harder than it looks. I did not find a satisfying solution.

### DIFF
--- a/alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
+++ b/alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
@@ -9,8 +9,6 @@ import com.gu.scanamo.query._
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.time.{Millis, Seconds, Span}
 import org.scalatest.{BeforeAndAfterAll, FunSpecLike, Matchers}
-import cats.data.NonEmptySet
-import cats.kernel.instances.all._
 
 class ScanamoAlpakkaSpec
   extends FunSpecLike
@@ -119,6 +117,7 @@ class ScanamoAlpakkaSpec
       case class Farm(asyncAnimals: List[String])
       case class Farmer(name: String, age: Long, farm: Farm)
 
+      import com.gu.scanamo._
       import com.gu.scanamo.syntax._
 
       val dataSet = Set(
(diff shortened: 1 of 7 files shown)


regiskuckaertz:
An alternative would be to provide a utility function in the `scanamo` package object, e.g.
```scala
def keySet[A](h: A, t: A*): Set[A] = t.toSet + h
```
Then the non-empty constraint becomes opt-in.

philwills:
I think this is a great idea and like the preclusion of error, but I'd be tempted by the opt-in approach initially.

regiskuckaertz:
Yeah you're right, I've updated.

philwills:
I'm in to minds about this. I'm not sure that this is valuable enough to warrant the extra explanation required.

regiskuckaertz:
I see what you mean. I still think there's value in having rules in scanamo to prevent such error cases, as they seem so trivial, and it relates to the mantra of functional programming that "well-formed terms don't get stuck". However there is no way around changing the type of:
```scala
    implicit def symbolListTupleToUniqueKeys[V: DynamoFormat](pair: (Symbol, Set[V])) =
      UniqueKeys(KeyList(pair._1, pair._2))

    implicit def toMultipleKeyList[H: DynamoFormat, R: DynamoFormat](pair: (HashAndRangeKeyNames, Set[(H, R)])) =
      UniqueKeys(MultipleKeyList(pair._1.hash -> pair._1.range, pair._2))
```
I'm happy to close this one and give it another thought for v2, what do you think?

regiskuckaertz:
Let's rethink that from a blank slate.

Issue #201, opened 2018-04-04
Optimise derivation for reduced compile time


regiskuckaertz:
Via semi-automatic derivation and/or magnolia.

Issue #168, opened 2017-11-09
Unable to create DynamoFormat when using scalaz tagged types


kvnflm:
Here's a minimal example that seems like it should compile, but does not:
```
import com.gu.scanamo.DynamoFormat
import com.gu.scanamo.DynamoFormat._
import scalaz.{@@, Tag}
import scalaz.syntax.tag._

sealed trait MyTag

case class TestClass(s: String @@ MyTag)

object DynamoFormatters {

  implicit val tagFormatter: DynamoFormat[String @@ MyTag] =
    DynamoFormat.iso[String @@ MyTag, String](Tag[String, MyTag])(_.unwrap)

  implicit val testClassFormatter = DynamoFormat[TestClass]
}
```

It fails with `could not find implicit value for parameter instance: com.gu.scanamo.DynamoFormat[TestClass]`. This is with scala 2.12.4, scanamo 1.0.0-M2, and scalaz-core 7.2.16.

On the other hand, I did notice that when using shapeless's version of `@@`, it does compile. The most obvious difference I see is that shapeless `@@` is covariant on the base type, while scalaz `@@` is invariant -- I'm not sure if this is relevant, though.

kvnflm:
As an update, I've switched to explicitly defining my `DynamoFormat`'s `read` and `write` instead of relying on the automatically derived versions to have a bit more control over how the data is written out. When doing so, the following works without problem:
```
implicit def taggedFormatter[BASE: DynamoFormat, TAG]: DynamoFormat[BASE @@ TAG] =
    DynamoFormat.iso[BASE @@ TAG, BASE](Tag[BASE, TAG])(_.unwrap)
```

So I suppose it's not accurate to say that Scalaz tagged types don't work with `DynamoFormat` as they work fine here. Either something with the auto-derivation was getting tripped up, or perhaps I was misusing it my first example -- feel free to close if this is actually a non-issue.

Issue #135, opened 2017-08-24
disable automatic derivation


fommil:
My team prefers semi-auto derivation as per https://github.com/fommil/stalactite

It would be great if you followed the pattern for derivation laid out in https://github.com/milessabin/export-hook to

1. avoid lots of corner cases you may not have considered
2. allow users to opt-in to full generic derivation

thanks for a great library!

philwills:
It's something I've discussed before and it's definitely the approach I would try and take were I starting from scratch. 

I'm not planning to work on it myself imminently, but I'd be open to a PR and discussing a migration approach.

fommil:
stalactite dramatically speeds up compiletimes, btw... we can talk about it at scala world if you're going?

philwills:
Would love to, but I've yet again failed to organise myself to go.

Baccata:
My team has been bitten a couple times by full-auto derivation, and we would love to see the derived formats made orphan.

mattroberts297:
It's a shame that we cannot share some of the fun examples (and less fun work arounds) actually. Totally understand that it's not something you have time to implement at the moment though :smile:.

Issue #138, opened 2017-09-07
Enum value types mapped incorrectly when used directly in queries


ivashin:
When an enumeration is defined as sealed trait + case objects hierarchy, using enum values directly in update operations results in record being corrupted because of incorrect mapping. Sample below fails with `PropertyReadError(status,TypeCoercionError(java.lang.Exception: {M: {},} is not a recognised member of the Enumeration))` error.

```scala
object EnumTest {
  sealed trait Status
  case object Pending extends Status
  case object Completed extends Status

  case class Record(id: String, status: Status)
}

class EnumTest extends FunSpec with Matchers {
  import EnumTest._

  lazy val client = AmazonDynamoDBClientBuilder.standard().withEndpointConfiguration(new EndpointConfiguration("http://localhost:8000", "")).build()
  lazy val table = Table[Record]("TestTable")

  describe("Scanamo") {
    val id = "ABC"
    val putRequest = table.put(Record(id, Pending))
    Scanamo.exec(client)(putRequest)

    it("should update enum value") {
      val updateRequest = table.update('id -> id, set('status -> Completed))
      val updated = Scanamo.exec(client)(updateRequest)
      updated.right.value should equal(Record(id, Completed))
    }
  }
}
```

Similarly, enum values can't be used directly in conditional clauses.

When enum is casted to base trait, either by `.asInstanceOf[Status]` or `val status: Status = `, everything works as expected.

philwills:
#140 should stop it accidentally compiling. Rather than using an `asInstanceOf`, you could write

```scala
val updateRequest = table.update('id -> id, set[Status]('status -> Completed))
```
which is a bit less ugly. 

Still not ideal, but I'm unsure how to improve this with this encoding of an enumeration without changing the variance of `DynamoFormat`, which is a bit too scary for my liking.

Down the line it would be good to do something like https://github.com/kailuowang/henkan does with it's `set` method to exploit the fact that the `Table` knows the types of its fields, but that feels a pretty chunky change.

PR #215, opened 2018-05-14
Semi-automatic derivation with Magnolia

regiskuckaertz:
Addresses #135, #201 and perhaps #168 too.

This PR substitutes [Magnolia](http://magnolia.work/) for Shapeless in the derivation of the `DynamoFormat` and `EnumerationDynamoFormat` typeclasses. It's still rough around the edges but it's getting there. To derive a format, users must now write, for type `T`:
```scala
implicit val format: DynamoFormat[T] = DynamoFormat.derive // if T is not an enumeration
implicit val format: EnumerationDynamoFormat[T] = DynamoFormat.deriveEnum // otherwise
```

I'm not familiar enough with the [export-hook](https://github.com/milessabin/export-hook) thingie so will need to do some reading.

TODO
- [x] Fix tests
- [x] Update scaladocs
- [x] Update tuts


regiskuckaertz:
- I think automatic derivation should still be available somewhere, ala Circe
- Regarding package organisation, perhaps offering package objects `semiauto` and `auto` then
- I'm wondering if the enumeration derivation is still necessary, both `combine` and `dispatch` seem general enough

philwills:
> I think automatic derivation should still be available somewhere, ala Circe
👍 

> I'm wondering if the enumeration derivation is still necessary, both combine and dispatch seem general enough
This needs a lot of care, as I had thought that previously and ran into #138. I haven't looked at this carefully enough to know, but it's worth bearing in mind.

philwills:
It would be great to know whether there's a noticeable compilation speed-up for a deeply ADT.

In the documentation, I'd err towards putting the derived format in the companion object, as instinctively that would be my preferred pattern. Happy to be argued out of that though.

regiskuckaertz:
> This needs a lot of care, as I had thought that previously and ran into #138.

Yeah, I think I found a way. Magnolia provides an `isObject` method on `CaseClass` and it provides a nice solution to the problem.

Now two tests are failing and I'm not sure why yet.

alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
```diff
@@ -145,7 +147,7 @@ class ScanamoAlpakkaSpec
       import com.gu.scanamo.syntax._
 
       val forecasts = for {
-        _ <- ScanamoAlpakka.update(alpakkaClient)("forecast")('location -> "London", set('weather -> "Sun"))
+        _ <- ScanamoAlpakka.update[Forecast](alpakkaClient)("forecast")('location -> "London", set('weather -> "Sun"))
```
regiskuckaertz: Not adding the type generates a diverging implicit error, but I see this is used in other places so assuming that's not too bad.

regiskuckaertz:
Automatic derivation for any `A` supplanted the derivation for `Some[A]` in `DynamoFormat`. That's where the export-hook thing comes in... is this a caveat we're willing to add to the documentation? Speaking of, I need to update the pages on custom formats...

Otherwise, looking good?

alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
philwills: I'm OK with it in this case. I'd be a lot more worried if it was one of the methods on `Table`. Just need to think about how to make it clear in release notes.

philwills:
This does look really good generally. I'm not quite sure what you mean with `Some[T]`, but `Option` is such a common type that any regression in it's ergonomics would be an issue.

regiskuckaertz:
> I'm not quite sure what you mean with Some[T]

The failing tests occurred when a column was of type `Option[T]` and the program tried to put `Some(x)` into it. There is a rule in the `DynamoFormat` object for that specific case, but it was foreshadowed by bringing `generic.auto._` into scope.

philwills:
Got it. Having the auto-derivation as part of the class hierarchy made it relatively simple to push it's priority down the list, but having it imported trumps the precedence of the companion object.

My usual tactic in these situations is to look what Circe has done and blatantly rip it off. https://github.com/circe/circe/pull/247/files looks like the most relevant diff, but there's a lot going on there to pick out the relevant parts from.

regiskuckaertz:
haha, good ideas travel far in the open source world 😄 

ok I see what he's doing, though I don't understand yet how it's working wrt priority. I'll try and see how I can reconcile this with Magnolia's macro. My first attempt (just wrapping an instance in `Exported` at the calling site) did not work.

regiskuckaertz:
oh I get it. The implicit def in `LowPriority` expects an implicit `Exported[DynamoFormat[A]]`, but that can only exist if you bring `io.circe.generic.auto._` in scope where it's being defined.

fommil:
great! Please add a `deriving.conf` using this format https://gitlab.com/fommil/scalaz-deriving/#deriving so that it works out of the box with `@deriving`

e.g. https://gitlab.com/fommil/scalaz-deriving/blob/master/examples/xmlformat/src/main/resources/deriving.conf

You could even add unit tests for this.

regiskuckaertz:
@fommil ooh nice 👍just did.

regiskuckaertz:
@philwills there's current activity in the magnolia project on that very issue (automatic derivation + scope), hopefully we can get that sorted soon.

dmitry-avershin:
what is the state of this one? would it be part of the upcoming release?

regiskuckaertz:
No, it won't go in, because fully automatic derivation subsumes implicits for concrete types, which I haven't found a solution to yet.
We're working on semi+full auto derivation within the existing shapeless code and that is going to ship in v1.

regiskuckaertz:
I expect scala 3 will be out before the auto derivation issue is solved

PR #217, opened 2018-05-19
Add derived format documentation

mattroberts297:
Because it is useful to understand the implementation before attempting
to use this awesome feature (microsite tested locally).


philwills:
Thank you. Great to see a practical example, rather than my flights of whimsy.

PR #218, opened 2018-05-19
Add ruby version

mattroberts297:
So that contributing to documentation is easier.

This is because the microsite plugin has a secret dependency on jekyll
and the latest version of jekyll is incompatible with the version of ruby
shipped on macOS. The only real competitor to rbenv is rvm and I don't
think this change is incompatible with that. With this file and rbenv
installed contributers can run gem install jekyll and things will mostly
just work.


philwills:
Great point. Thanks.

Issue #211, opened 2018-05-09
Readme example doesn't work for me.


RyanSattler:
LocalDynamoDB is not found in any of the Scanamo jars that come from importing "com.gu" %% "scanamo" % "1.0.0-M6" . Not sure if there is a dependency missing.

mattbernst:
Code has changed and docs are outdated. See comments on this closed issue for a workaround: https://github.com/scanamo/scanamo/issues/189

regiskuckaertz:
Hi. You can know use LocalDynamoDB in the `testkit` sub-project.

Issue #204, opened 2018-04-04
Set up test infrastructure


regiskuckaertz:
Make it possible to use things like LocalDynamoDB outside of Scanamo itself to simplify testing where you wanted to actually probe the real API.

rtyley:
We wanted to test in Ophan that we were persisting a field correctly (DynamoDB TTL demands you store the expiration time as epoch seconds, not epoch milliseconds, d'oh!) and could really have done with some of the nice ARM methods in `com.gu.scanamo.LocalDynamoDB`.

Note also that there are two copies of this class:

* alpakka/src/test/scala/com/gu/scanamo/LocalDynamoDB.scala
* scanamo/src/test/scala/com/gu/scanamo/LocalDynamoDB.scala

cc @amyhughes

Issue #189, opened 2018-02-15
where is LocalDynamoDB?


poohsen:
The basic usage example mentions LocalDynamoDB but I don't see it in the master branch nor in the 1.0.0-M2, M3 or M4 builds. Where does this object come from? Feels like I'm missing something major... a little help would be much appreciated

mattbernst:
The examples are broken because LocalDynamoDB isn't accessible that way any more. If you want to look at it, see: https://github.com/scanamo/scanamo/blob/master/scanamo/src/test/scala/com/gu/scanamo/LocalDynamoDB.scala

If you just want to run the examples against a local dynamo db instance, you can do something like

```
val client = {
    val conf = new EndpointConfiguration("http://localhost:8000", "us-east-1")
    AmazonDynamoDBClientBuilder.standard().withEndpointConfiguration(conf).build()
  }
```

wherever the examples currently show `val client = LocalDynamoDB.client()`.

poohsen:
that worked. thanks!

PR #219, opened 2018-05-19
Split off LocalDynamoDB into 'testing' module for publishing

rtyley:
This commit addresses https://github.com/scanamo/scanamo/issues/204 - making `LocalDynamoDB` available in a published artifact, which might help with the confusion seen in https://github.com/scanamo/scanamo/issues/189 &  https://github.com/scanamo/scanamo/issues/211.


philwills:
This looks good, though I think `testkit` is probably a more common name than `testing`.

rtyley:
> This looks good, though I think testkit is probably a more common name than testing.

Redone as 'testkit' with b30a15a !

Issue #224, opened 2018-06-11
Is the `IN` operator supported in Query?


rakshith-oath:
DynamoDb Java SDK supports `IN` comparison operator (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html#Expressions.OperatorsAndFunctions.Comparators).

Is this operator supported in Scanamo?
If not, can we please add the support for it?

Thanks!

regiskuckaertz:
Hello, no not yet. I've recently come across a query that would be much less expensive with `IN`, so I plan to add it.

regiskuckaertz:
This is now supported as of M7

PR #225, opened 2018-06-20
Add support for IN operator

regiskuckaertz:
Addresses #224 

I've just fallen victim of an N+1 query, that was killing the performance of my lambda. Using the `IN` operator reduces running time by two orders of magnitude.

### DIFF
--- a/scanamo/src/main/scala/com/gu/scanamo/query/ConditionExpression.scala
+++ b/scanamo/src/main/scala/com/gu/scanamo/query/ConditionExpression.scala
@@ -75,7 +75,7 @@ object ConditionExpression {
       RequestCondition(
         s"""#${attributeName.placeholder(prefix)} IN ${attributeValues.keys.mkString("(", ",", ")")}""",
         attributeName.attributeNames(s"#$prefix"),
-        Some(attributeValues.toMap)
+        Some(attributeValues)
       )
     }
   }
--- a/scanamo/src/main/scala/com/gu/scanamo/query/ConditionExpression.scala
+++ b/scanamo/src/main/scala/com/gu/scanamo/query/ConditionExpression.scala
@@ -58,6 +58,28 @@ object ConditionExpression {
     }
   }
 
+  implicit def symbolValueInCondition[V: DynamoFormat] = new ConditionExpression[(Symbol, Set[V])] {
+    override def apply(pair: (Symbol, Set[V]))(condition: Option[RequestCondition]): RequestCondition = {
+      attributeValueInCondition.apply((AttributeName.of(pair._1), pair._2))(condition)
+    }
+  }
+
+  implicit def attributeValueInCondition[V: DynamoFormat] = new ConditionExpression[(AttributeName, Set[V])] {
+    val prefix = "inCondition"
+    override def apply(pair: (AttributeName, Set[V]))(condition: Option[RequestCondition]): RequestCondition = {
+      val format = DynamoFormat[V]
+      val attributeName = pair._1
+      val attributeValues = pair._2.zipWithIndex.map { case (v, i) => 
+        s":conditionAttributeValue$i" -> format.write(v)
+      }.toMap
+      RequestCondition(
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
Compilation would fail with something like
```scala
table.filter('xyz -> "abc" and 'mno -> Set(1,2,3))
```
but then this works
```scala
table.filter(AndCondition('xyz -> "abc", 'mno -> Set(1,2,3)))
```
The failure happens because the `and` operator is overloaded in at least two different settings, and the wrong one wins. There's probably a way around that but it will require more groundwork with the typeclasses.

scanamo/src/main/scala/com/gu/scanamo/query/ConditionExpression.scala
```diff
@@ -58,6 +58,28 @@ object ConditionExpression {
...
+  implicit def attributeValueInCondition[V: DynamoFormat] = new ConditionExpression[(AttributeName, Set[V])] {
+    val prefix = "inCondition"
+    override def apply(pair: (AttributeName, Set[V]))(condition: Option[RequestCondition]): RequestCondition = {
+      val format = DynamoFormat[V]
+      val attributeName = pair._1
+      val attributeValues = pair._2.zipWithIndex.map { case (v, i) => 
+        s":conditionAttributeValue$i" -> format.write(v)
+      }.toMap
+      RequestCondition(
+        s"""#${attributeName.placeholder(prefix)} IN ${attributeValues.keys.mkString("(", ",", ")")}""",
```
hosamaly: Should the values be always quoted? Even if they aren't strings?

hosamaly: If `attributeValues.isEmpty` then we'll get a query that checks for an empty string (`attr IN ("")`). Is this the desired behaviour?

regiskuckaertz: Given a pair `'hello -> Set(1,2,3)`, this will produce `#inConditionhello IN (:conditionAttributeValue0,:conditionAttributeValue1,:conditionAttributeValue2)`. In the DynamoDB API, `#str` is a name binding expression, and `:str` is a value binding expression.

- for keys: `query.withExpressionAttributeNames(Map("#str" -> "UserId"))`
- for values, you must bind each expression to a corresponding `AttributeValue`, e.g.: `query.withExpressionAttributeValues(Map(":str" -> new AttributeValue().withN("1234")))`

regiskuckaertz: Yes, that is entirely possible. It is a consequence of using `Set` instead of `NonEmptySet`. I have tried to address this [previously](https://github.com/scanamo/scanamo/pull/2130) but decided to delay this until v1 is out.

manjotmona:
hey @regiskuckaertz I am trying to apply a filter with query operation like this:

 val queryOperation = userIndex
            .filter(userID` -> Set(userIDs)
            .query(userName` -> "abc")

Here userName is a GSI and I want to further filter my results on userID. 

But this is giving me a compile-time error:

could not find implicit value for evidence parameter of type com.gu.scanamo.query.ConditionExpression[(Symbol, Set[ " "])]

I have tried this also:
 val queryOperation = userIndex
            .filter(AndCondition((userName` -> "abc", userID` -> Set(userIDs))
            .query(userName` -> "abc")

But it gives the same error.

regiskuckaertz:
Hi @manjotmona 

What is the type of `userIDs`? If it is a set of something (e.g. `Set[String]`), you don't need to wrap it with `Set(...)`. Another small thing is that you are already providing the key condition in `.query(...)` so you don't need to repeat it in the filter expression.

What about this:
```scala
val queryOperation = userIndex.filter('userID -> userIDs).query('userName -> "abc")
```

If `userIDs` is another type of collection, `userIDs.toSet` should get you there.

manjotmona:
Hi @regiskuckaertz, 
It is a set of enum.
And this time i dint wrap it in Set but it is giving me same error.

Issue #223, opened 2018-05-29
Ability for clients to page through query results


shayangz:
There are situations where results of a query can be quite large and clients would benefit from processing results in a paginated manner. dynamodb supports this type of experience by including a `LastEvaluatedKey` value in query response allowing subsequent queries to pickup where the last page ended by passing in a `ExclusiveStartKey` parameter (see [docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html)).

Exposing this information at the scanamo API level would greatly benefit clients using the library to be able to make appropriate paginated calls.

shayangz:
@regiskuckaertz, thanks for taking this on. As part of this work, it would be fantastic to also allow the DSL to set the `ScanIndexForward` option as well. Doing so would allow clients to do pagination in either direction.

regiskuckaertz:
Closed by #227 

(Note: `ScanIndexForward` is supported on queries: `table.query(...).descending`)

Issue #41, opened 2016-05-25
Support manual pagination keys


williamho:
Currently there's no way to manually specify pagination params when querying -- would be nice to have some way of getting the lastEvaluatedKey of a query, and being able to pass an exclusiveStartKey.

avpatel257:
I believe this is a very common use case where a query returns many records and clients need a way to paginate through it. It would be nice if we can pass the paging params (or last evaluated key) to dynamo and have dynamo only return results specified by limit param.

philwills:
I did consider this when implementing limit support, but was struggling to think of a good use case for this. Could you describe what you'd want it for in more detail?

Would it be enough to be able to provide the next batch of results for a given batch of results?

williamho:
mostly for client-side pagination, where the client only needs to display few items at a time, and a way to resume from where they left off (in my case, since the client can't pass in a `java.util.Map[String, AttributeValue]`, I'd encode/decode it as a  `String`)

in my own code, I ended up just doing something like:
```scala
case class DynamoPage[T](
  items: Seq[T],
  lastEvaluatedKey: java.util.Map[String, AttributeValue]
)

val table: Table[Something] = ???

def queryPaginated[T: DynamoFormat](
  query: Query[_],
  limit: Option[Int],
  startKey: Option[java.util.Map[String, AttributeValue]]
): ScanamoOps[DynamoPage[T]] = {
  val req = new QueryRequest().withTableName(table.name)

  limit.foreach(req.setLimit(_))
  startKey.foreach(key ⇒ req.setExclusiveStartKey(key))

  ScanamoOps.query(query(req)).map { qr: QueryResult ⇒
    val items = qr.getItems.asScala.map(ScanamoFree.read[T])
    val key = Option(qr.getLastEvaluatedKey)

    DynamoPage(items, key)
  }
}
```

that's a simplified version to get the gist of it, though the actual code I have is more complicated:
* has a typeclass `PageKeyFormat` to convert between `j.u.Map[String, AttributeValue]` and a type `K` (Key)
* the lastEvaluatedKey in `DynamoPage` is actually an `Option[K]`
* I define a `PageKeyFormat` to convert between the java Map to a compressed JSON string, which is the string key that the clients pass in to specify their `lastEvaluatedKey`

JacopoGobbi:
@williamho could you be able to provide some extra details on how to use your example outside of scanamo? Do you define your own interpreter?

williamho:
[...]

PR #227, opened 2018-06-23
Add pagination API

regiskuckaertz:
We've got #223 and and #41 making a case for manual pagination. I think this is a valid use case, in particular all interactive systems displaying lists of results eventually need this for obvious usability reasons—e.g. through pagination or lazy lists.

This PR introduces four new primitives: `scanFrom`, `scanIndexFrom`, `queryFrom` and `queryIndexFrom`. They differ from their `withLimit` counterpart in two ways:
- accept an optional parameter `startKey` used to offset the query
- returns a tuple where the left part is the list of results and the right part the last key in case not all available results have been returned.

- [x] Add tests
- [ ] External/Internal `EvaluationKey` (with a better name)
- [x] Update docs/tuts


philwills:
Would adding an additional method, say `scanFrom` or something along those lines work and cause less API disruption?

Also, I am struggling to keep up with the absolute tear you are on Regis 👏 👏 👏

regiskuckaertz:
I think that's a good idea. Then we can listen for users' feedback for a future version.

There's loads to come Phil 😅

regiskuckaertz:
@philwills if you're ok with it, I'd like to get this one merged, and then improve the interface in a follow-up PR, as part of https://github.com/scanamo/scanamo/issues/202

PR #230, opened 2018-07-07
Fix scalaz interpreter

regiskuckaertz:
Finally got around to fixing these tests. Conditional operations reify corresponding exceptions in the happy path, only those operations needed adjusting.

cc @emilypi


regiskuckaertz:
Those build failures are 😲

philwills:
Sorry for not reviewing. I wonder if your work on the build means that some of the extra shutdown calls can now be removed.

regiskuckaertz:
They're not fundamental indeed, I just thought a test suite creates a client and then shuts it down.

regiskuckaertz:
@emilypi We need to design a proper error ADT, `ConditionalCheckFailedException` only fits a few use cases and is a throwable. Then I totally agree we'll be able to use the full expressiveness of zio.

In hindsight, I could have fixed it by just `attempt`ing the conditional operations as they expect an `Either[ConditionalCheckFailedException, A0]` in return.

PR #232, opened 2018-07-08
Fix build

regiskuckaertz:
Instead of running Dynamo at task boundaries, it seems starting the server once before the aggregate `test` task is more efficient and also more reliable. I also removed a lot of redundancy and fixed a few bugs in the tests. Given snippets in scaladocs run in parallel (this is made evident by the order in which they are displayed), it is not recommended to use `LocalDynamoDB.createTable` and replace it with `withTable` where possible. I had to remove a couple of snippets in the scaladoc of `Scanamo.query`, I don't think that's too bad.


philwills:
In general this looks good, but it makes me a bit sad that I think this means you can't just run makeMicrosite without manually starting and stopping local dynamo. 

Can't confirm as my laptop is currently hosed.

regiskuckaertz:
@philwills hello 👋I can definitely improve that, I agree that's not very practical!

PR #239, opened 2018-07-22
Random table API in testkit

regiskuckaertz:
Coming up with table names is an annoyance in tests, only exacerbated by the fact that there may be some race condition lurking around (explaining the "cannot create preexisting table" exceptions occurring on Travis but not locally).

This PR introduces three new APIs in the `LocalDynamoDB` facility:

- `withRandomTable`
- `withRandomTableWithSecondaryIndex`
- `usingRandomTable`

They work exactly the same as their non-random counterparts, only that the thunk takes the table name as input.

### DIFF
--- a/alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
+++ b/alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
@@ -226,11 +226,11 @@ class ScanamoAlpakkaSpec extends FunSpecLike with BeforeAndAfterAll with Matcher
   it("scanIndexWithLimit") {
     case class Bear(name: String, favouriteFood: String, alias: Option[String])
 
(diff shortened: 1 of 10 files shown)


philwills:
It always feels a bit of a hack to rely on randomness to avoid this kind of conflict, but the table names were always just noise, so :+1:

regiskuckaertz:
@philwills there seems to be some race condition in tests, as since I changed the build "cannot create preexisting table" exceptions are thrown from time to time. It may just be SBT tests scheduling, I'm not sure—requires more investigation. In the meantime, this PR solves the problem by create tables with random names 😄

regiskuckaertz:
Yeah, I agree. On the flip side, I'm happy this change surfaced a behaviour that may be a bug somewhere; it may be a silly thing like an async test leaking resources. I'll sit and go through each test when I've got time.

PR #248, opened 2018-08-26
Add `Table.queryFrom` and `Table.scanFrom`

regiskuckaertz:
Adds the ability to paginate using the `Table` API, as methods other than `exec` on `Scanamo*` will eventually be [deprecated](#146).

At the moment, it requires the addition of two new APIs on `TableWithOptions`, namely `scanWithLastKey` and `queryWithLastKey`, so that the result can contain the last pagination key if there is one. This is not ideal, augmenting the surface API introduces new opportunities for bugs to creep in. What we want are "all-powerful" scan/query operations, let's call them `scan0`/`query0` and all the other variants piggyback on them with ad hoc defaults.

Another issue I need to figure out is the asymmetry between the input key (a `UniqueKey[_]`) and the output key (a `java.util.Map[String, AttributeValue]`). I have added enriched `UniqueKeyCondition` with a `fromAVMap` function, but it is not yet clear when or where to call it; I still need to look into that.

### DIFF
--- a/scanamo/src/main/scala/com/gu/scanamo/Table.scala
+++ b/scanamo/src/main/scala/com/gu/scanamo/Table.scala
@@ -510,7 +510,7 @@ case class Table[V: DynamoFormat](name: String) {
   def scan(): ScanamoOps[List[Either[DynamoReadError, V]]] = ScanamoFree.scan[V](name)
 
   /**
-    * Scans all elements of a table
+    * Scans all elements of a table starting from the specified key
     *
     * {{{
     * >>> case class Bear(name: String, favouriteFood: String)
@@ -566,7 +566,7 @@ case class Table[V: DynamoFormat](name: String) {
   def query(query: Query[_]): ScanamoOps[List[Either[DynamoReadError, V]]] = ScanamoFree.query[V](name)(query)
 
   /**
-    * Query a table based on the hash key and optionally the range key
+    * Query a table based on the hash key and optionally the range key, starting from the specified key
     *
     * {{{
     * >>> case class Transport(mode: String, line: String)
--- a/scanamo/src/main/scala/com/gu/scanamo/Table.scala
+++ b/scanamo/src/main/scala/com/gu/scanamo/Table.scala
@@ -293,34 +293,6 @@ case class Table[V: DynamoFormat](name: String) {
   def limit(n: Int) = TableWithOptions[V](name, ScanamoQueryOptions.default).limit(n)
 
   /**
-    * Query or scan a table, starting from a particular item identified by its key
-    * {{{
-    * >>> case class Transport(mode: String, line: String)
-    *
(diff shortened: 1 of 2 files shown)


scanamo/src/main/scala/com/gu/scanamo/Table.scala
```diff
@@ -508,6 +509,34 @@ case class Table[V: DynamoFormat](name: String) {
     */
   def scan(): ScanamoOps[List[Either[DynamoReadError, V]]] = ScanamoFree.scan[V](name)
 
+  /**
+    * Scans all elements of a table
```
philwills: Need to document what from means.

scanamo/src/main/scala/com/gu/scanamo/Table.scala
```diff
@@ -536,6 +565,38 @@ case class Table[V: DynamoFormat](name: String) {
     */
   def query(query: Query[_]): ScanamoOps[List[Either[DynamoReadError, V]]] = ScanamoFree.query[V](name)(query)
 
+  /**
+    * Query a table based on the hash key and optionally the range key
```
philwills: Docs

regiskuckaertz:
I agree. Initially that's what I had, but indeed there's no way around changing the return type of `scan` and `query`. `List[Either[A, B]]` is `EitherT[List, A, B]`, so we can probably come up with a monad transformer of our doing that provides access to the last evaluated key, if any. Something to think about for a future one.

philwills:
Did you try `table.from(x).query(y)`? That would seem to avoid the problem and fit with the approach to `limit` and friends.

regiskuckaertz:
Yes and it worked for the first half (the input); however, because it returns a list, you lose access to the [last evaluated key](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/model/ScanResult.html#getLastEvaluatedKey--), rendering this version useless 😞 

I imagine:

```scala
case class ScanamoResultT[M[_], A] private (private val value: EitherT[M, DynamoReadError, A]) {
  type K
  def lastEvaluatedKey: Option[UniqueKey[K]]
  // ???
}
type ScanamoResult[A] = ScanamoResultT[Id, A]
```

Then we can update `Table`:
```scala
  def update(key: UniqueKey[_], expression: UpdateExpression): ScanamoOps[ScanamoResult[V]]
  def scan: ScanamoOps[ScanamoResultT[List, V]]
```
and get back the elegance of `table.limit(???).from(???).scan` with access to whatever we feel necessary.

regiskuckaertz:
ping

philwills:
Sorry for the silence. I think this is a good enough compromise that it makes sense to merge.

I should probably have named the original method `scanAll`, but we are where we are.

regiskuckaertz:
No worries, there is a way out, but we can think about it when v1 is released.

PR #250, opened 2018-08-27
[DynamoFormat] Read a Set which has been deleted after an update operation

### LINKED ISSUE #244: Update expression delete will turns a string set into a null value, which causes the DynamoFormat to break
Basically, the following example:
```
val client = LocalDynamoDB.client()
import com.amazonaws.services.dynamodbv2.model.ScalarAttributeType._
val farmersTableResult = LocalDynamoDB.createTable(client)("farmer")('name -> S)
case class Farmer(name: String, attrs: Set[String])

val farmers = Table[Farmer]("farmer")

val updateOps = for {
  _ <- farmers.put(Farmer("bob", Set("woo")))
  _ <- farmers.update('name -> "bob", delete('attrs -> Set("woo")))
} yield ()
val getOps = for {
  bob <- farmers.get('name -> "bob")
} yield bob

println(Scanamo.exec(client)(updateOps))
println(Scanamo.exec(client)(getOps))
```

Results in:
```
()
Some(Left(InvalidPropertiesError(NonEmptyList(PropertyReadError(attrs,MissingProperty)))))
```

So I think https://github.com/scanamo/scanamo/blob/master/formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala#L313 needs to be updated to treat a `null` string set as an empty set.

Full worked example here, https://github.bamtech.co/gist/jcollier1/8e09fa1fa44235723e99d5ec8e85b4f6 .

Slakah:
# TODO
- [ ] Fix other set types.

Fixes #244 .

### DIFF
--- a/formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala
+++ b/formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala
@@ -115,6 +115,7 @@ object DynamoFormat extends EnumDynamoFormat {
   def iso[A, B](r: B => A)(w: A => B)(implicit f: DynamoFormat[B]) = new DynamoFormat[A] {
     override def read(item: AttributeValue): Either[DynamoReadError, A] = f.read(item).map(r)
     override def write(t: A): AttributeValue = f.write(w(t))
+    override val default: Option[A] = f.default.map(r)
   }
 
   /**
@@ -310,7 +311,14 @@ object DynamoFormat extends EnumDynamoFormat {
     )(javaListFormat)
 
   private val javaNumSetFormat = attribute(_.getNS, "NS")(_.withNS)
-  private val javaStringSetFormat = attribute(_.getSS, "SS")(_.withSS)
+  private val javaStringSetFormat: DynamoFormat[java.util.List[String]] =
+    new DynamoFormat[java.util.List[String]] {
+      override def read(av: AttributeValue): Either[DynamoReadError, java.util.List[String]] =
+        Either.fromOption(Option(av.getSS), NoPropertyOfType("SS", av))
+      override def write(t: java.util.List[String]): AttributeValue =
+        new AttributeValue().withSS(t)
+      override val default: Option[java.util.List[String]] = Some(java.util.Collections.emptyList())
+    }
   private def setFormat[T](r: String => Either[DynamoReadError, T])(w: T => String)(
       df: DynamoFormat[java.util.List[String]]): DynamoFormat[Set[T]] =
     xmap[Set[T], java.util.List[String]](_.asScala.toList.traverse(r).map(_.toSet))(
@@ -371,7 +379,7 @@ object DynamoFormat extends EnumDynamoFormat {
     * }}}
     */
   implicit val stringSetFormat =
-    xmap[Set[String], java.util.List[String]](s => Right(s.asScala.toSet))(
+    iso[Set[String], java.util.List[String]](s => s.asScala.toSet)(
       _.toList.asJava
     )(javaStringSetFormat)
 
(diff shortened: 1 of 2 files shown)


formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala
```diff
@@ -115,6 +115,7 @@ object DynamoFormat extends EnumDynamoFormat {
   def iso[A, B](r: B => A)(w: A => B)(implicit f: DynamoFormat[B]) = new DynamoFormat[A] {
     override def read(item: AttributeValue): Either[DynamoReadError, A] = f.read(item).map(r)
     override def write(t: A): AttributeValue = f.write(w(t))
+    override val default: Option[A] = f.default.map(r)
```
regiskuckaertz: 👍

formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala
```diff
@@ -310,7 +311,14 @@ object DynamoFormat extends EnumDynamoFormat {
     )(javaListFormat)
 
   private val javaNumSetFormat = attribute(_.getNS, "NS")(_.withNS)
-  private val javaStringSetFormat = attribute(_.getSS, "SS")(_.withSS)
+  private val javaStringSetFormat: DynamoFormat[java.util.List[String]] =
```
regiskuckaertz: Not too keen on having instances for mutable types, but I see that was the direction chosen previously.

regiskuckaertz:
This is great ❤️ 

In the future, I'd prefer something like this:
```scala
private def setFormat[S, T](
  f: AttributeValue => java.util.List[S], g: java.util.List[S] => AttributeValue)(
  iso: Iso[S, T]): DynamoFormat[Set[T]]
```
and derive `stringSetFormat`, `numberSetFormat` and `binarySetFormat` from it. The `Iso` is because numbers are encoded as strings in DynamoDB and our instance wouldn't be lawful, but I can live with that.

regiskuckaertz:
In fact `javaNumSetFormat` is subject to the same behaviour, so will need updating too.

Slakah:
Sorry I should have mentioned that this is in progress. I'm going to be
moving house this next week (very busy), but should be able to progress on
the rest of this PR soon(tm).

Slakah:
Hi @regiskuckaertz 

I've updated the PR to fix the number set values as well.

I've also changed the set formatters to encode an empty set as a `null` value, due to dynamo not supporting the notion of empty sets:
> Set type attributes cannot be empty.
[source](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html)

This can be removed if required, as it's a tad more controversial, but it does feel like an `AttributeValue` wrote by a `DynamoFormat` should be insertable into dynamo.

wrt.
> and derive stringSetFormat, numberSetFormat and binarySetFormat from it. 

The reason this wouldn't be possible, is because the decoding of a number can fail, so at the very least an `xmap` would need to be used.

formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala
```diff
@@ -309,71 +310,102 @@ object DynamoFormat extends EnumDynamoFormat {
...
-  private def setFormat[T](r: String => Either[DynamoReadError, T])(w: T => String)(
-      df: DynamoFormat[java.util.List[String]]): DynamoFormat[Set[T]] =
-    xmap[Set[T], java.util.List[String]](_.asScala.toList.traverse(r).map(_.toSet))(
-      _.map(w).toList.asJava
-    )(df)
+  private def numSetFormat[T](r: String => Either[DynamoReadError, T])(w: T => String): DynamoFormat[Set[T]] =
+    new DynamoFormat[Set[T]] {
+      override def read(av: AttributeValue): Either[DynamoReadError, Set[T]] =
+        for {
+          ss <- Either.fromOption(Option(av.getSS), NoPropertyOfType("NS", av))
```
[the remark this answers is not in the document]
Slakah: Good spot! Turns out I managed to break the tests, which was hiding the fact that I wired this in wrong. Should have been `getNS` instead of `getSS` etc.

See: https://github.com/scanamo/scanamo/pull/250/commits/bab4151900ae6c724e491e541d9df8d9d1031555

philwills:
Late to this, but agree with @regiskuckaertz that this looks great. Thanks @Slakah

PR #258, opened 2018-09-18
Add scalafmt config

regiskuckaertz:
I've stolen the config from the scalaz-zio project, I find it strikes a nice balance.

Note that if a file is not properly formatted, the build will fail.


.scalafmt.conf
```diff
@@ -1,2 +1,14 @@
...
\ No newline at end of file
+align = most
+continuationIndent.defnSite = 2
+assumeStandardLibraryStripMargin = true
+docstrings = JavaDoc
+lineEndings = preserve
+includeCurlyBraceInSelectChains = false
+danglingParentheses = true
+spaces {
+  inImportCurlyBraces = true
```
philwills: I don't think this adds anything and makes this even more disruptive.

.scalafmt.conf
```diff
@@ -1,2 +1,14 @@
 maxColumn = 120
-align = none
\ No newline at end of file
+align = most
```
philwills: I'm in two minds on this one. I do think it makes some things clearer, but sometimes it makes things look related when they're not. It also significantly adds to the size of this PR.

regiskuckaertz: Can you provide an example? We can tweak that parameter [in many ways](https://scalameta.org/scalafmt/docs/configuration.html#alignment). I agree, in general formatting will always increase the size of a PR, I think this is an exceptional case though -- in a "stable" branch the impact will be minor.

philwills: Looking through again, none of them are that heinous.

One thing I really don't like on second viewing is moving multiline argument lists so that they start next to the opening parenthesis, rather than on s slightly indented new line. It moves a lot of information a long way from left alignment.

regiskuckaertz: Agreed, I've updated.

philwills:
Thanks for putting that extra into it, @regiskuckaertz

Issue #267, opened 2018-11-05
Support querying tables in descending order


jcohen:
By default, Dynamo returns results in ascending order. Dynamo also allows for retrieving results in descending order by sending `false` as the value for the `scanIndexForward` property on the query request.

regiskuckaertz:
Closing as discussed in #268

PR #268, opened 2018-11-05
Add an option to query tables in reverse order.

jcohen:
This should address #267. Please let me know if I missed adding support for this anywhere (e.g. does it need to be added to the top level Scanamo/ScanamoAsync/etc.), or if more tests are needed, etc.!

### DIFF
--- a/scanamo/src/main/scala/com/gu/scanamo/Table.scala
+++ b/scanamo/src/main/scala/com/gu/scanamo/Table.scala
@@ -293,6 +293,31 @@ case class Table[V: DynamoFormat](name: String) {
   def limit(n: Int) = TableWithOptions[V](name, ScanamoQueryOptions.default).limit(n)
 
   /**
+    * Query a table in reverse order
+    * {{{
+    * >>> case class Transport(mode: String, line: String)
+    *
+    * >>> val client = LocalDynamoDB.client()
+    * >>> import com.amazonaws.services.dynamodbv2.model.ScalarAttributeType._
+    *
+    * >>> LocalDynamoDB.withRandomTable(client)('mode -> S, 'line -> S) { t =>
+    * ...   import com.gu.scanamo.syntax._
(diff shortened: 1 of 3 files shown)


regiskuckaertz:
Hi @jcohen 

This is actually supported already, but the combinator is on the condition instead of the table:
```scala
table.query(('key -> value).descending)
```
The reason is that it only works for queries, not scans, and with this PR someone could do `table.reverse.scan(...)` even if it does not make sense.

From a usability POV, I think it makes sense to move that in the table API though. However we need to find a solution to prevent the kind of degenerate case as above. I see two:

1. create a `TableQueryOnlyOptions[V]` that only permits queries: there will be duplication but will get the job done
2. reverse the ordering in the expression, i.e. `table.query(...).reverse`. This requires more effort but is the right thing to do I believe. For example in this case, we could introduce a new primitive to `ScanamoOpsA`:

```scala
final case class Reverse(query: ScanamoOps[QueryResult]) extends ScanamoOps[QueryResult]
```
then add a case to the interpreters:
```scala
case Revere(query) =>
  val q = query.asInstanceOf[Query]
  val r = q.req.withScanForward(false)
  apply(Query(r))
```
How does that sound?

regiskuckaertz:
I'm tempted to close this PR for now until we get 1.0 out, after which we can start thinking of an API overhaul for v2.

jcohen:
Sorry for the delayed response.

@regiskuckaertz I think that's fair. The existing functionality (using `.descending` on the key) works well enough for now.

I'm not sure how I missed it to begin with. I swear I grepped the codebase for `scanIndexForward` before throwing this together ;).

toidiu:
I have a scenario where I have a partition key(P) and a sort key(S). I want the query:
`('P = value) and ('S -> descending)`

Am I missing something or is there a way to do this with the `.descending` syntax.

PR #269, opened 2018-11-06
Use Async rather than Effect in cats-effect

tomverran:
Hi,

This PR just find/replaces `Effect` with `Async` as there are no uses of `runAsync` or `toIO` and I want to use a `StateT` as my effect type, which can only have an `Async` instance.

Hope this looks okay, let me know if you'd like any changes (or if I've missed something about `Effect`)

... actually it looks like I'm going in the wrong direction by trying to use `StateT`, but I guess this PR still has relevance. Looks like the tests just failed due to a timeout but I can't rerun them :(

### DIFF
--- a/cats/src/main/scala/com/gu/scanamo/ScanamoCats.scala
+++ b/cats/src/main/scala/com/gu/scanamo/ScanamoCats.scala
@@ -1,6 +1,6 @@
 package com.gu.scanamo
 
-import cats.effect.Effect
+import cats.effect.Async
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsync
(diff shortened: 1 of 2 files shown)


LATaylor-guardian:
> (It may just be that you have to write a LiftIO for your monad stack, I don't know if cats-effect provides them for you ...) 

https://typelevel.org/cats-effect/typeclasses/liftio.html

dmitry-avershin:
@tomverran thank you very much for fixing that, we were actually bitten by that. @regiskuckaertz would it be  possible to make a new release soon that incorporates this change? that would be awesome, thx

regiskuckaertz:
@dmitry-avershin I'm in the run-up to get 1.0 out in the next month

PR #273, opened 2018-11-16
Remove all deprecated methods

regiskuckaertz:
As 1.0 is nearing release, I think we can take a bold step and get rid of all deprecated methods. 

I had already written the ZIO tests using the table API so it was just a matter of copy-pasting and changing the interpreter 🎉 

What do you think @philwills ?


regiskuckaertz:
@TBonnin Great idea! See #276

regiskuckaertz:
I wouldn't mind a review 🤗

alpakka/src/test/scala/com/gu/scanamo/ScanamoAlpakkaSpec.scala
```diff
@@ -178,320 +184,300 @@ class ScanamoAlpakkaSpec extends FunSpecLike with BeforeAndAfterAll with Matcher
...
+  //         res3 <- bears.index(i).limit(1).scanFrom(res2._2)
+  //       } yield res2._1 ::: res3._1
+  //     } yield bs
+
+  //     ScanamoAlpakka.exec(alpakkaClient)(result).futureValue should equal(
+  //       List(Right(Bear("Yogi", "picnic baskets", Some("Kanga"))), Right(Bear("Pooh", "honey", Some("Winnie"))))
+  //     )
+  //   }
+  // }
+
```
TBonnin: why is this commenting out?

regiskuckaertz: good question. they were not working as expected with the `scanFrom` API; I will uncomment them once I've rebased [the PR](#274) that introduces `from` and `query0` 🤗

regiskuckaertz:
Thank you!

PR #278, opened 2018-12-18
Add automatic and semi-automatic derivation of DynamoFormat

### LINKED ISSUE #201: Optimise derivation for reduced compile time

TBonnin:
- Auto derivation of case classes now requires `import com.gu.scanamo.auto._`
and doesn't come automatically as soon as DynamoFormat is in scope
- Adding semi-automatic derivation via `deriveDynamoFormat[A]` helper method

Closes #201

### DIFF
--- a/formats/src/main/scala/com/gu/scanamo/export/Exported.scala
+++ b/formats/src/main/scala/com/gu/scanamo/export/Exported.scala
@@ -1,0 +1,3 @@
+package com.gu.scanamo.export
+
+final case class Exported[T](instance: T) extends AnyVal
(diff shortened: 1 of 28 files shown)


formats/src/main/scala/com/gu/scanamo/export/Exported.scala
```diff
@@ -0,0 +1,3 @@
+package com.gu.scanamo.export
+
+case class Exported[T](instance: T) extends AnyVal
```
regiskuckaertz: Let's make if `final` too

regiskuckaertz:
Yoohoo, thanks for your first contribution @TBonnin ✌️

PR #280, opened 2018-12-19
New build pipeline

### LINKED ISSUE #281: Failing tests in 2.11
Loads of tests are failing in 2.11. The root of the issue seems to be how we deal with `NULL` in the derivation. This must be fixed before we release v1.

regiskuckaertz:
Closes #281 

Many people have asked when the next release will happen. In the run-up to v1, we've been adding new features, changing the API, etc. so understandably users want the very latest version. I think this is something that will happen again and again when we approach deadlines.

To improve on our existing setup, this change introduces two drastic changes:
- every merge on the master branch will trigger a snapshot release in Sonatype, that will look something like `scanamo-<commitId>`
- releases will be triggered by Git tags, rather than the ol' Sonatype process. This is a long pet-peeve of mine that we don't use Git workflows that much, even when it's built for that.


regiskuckaertz:
@philwills brought back coverage analysis (without the pesky comments) 🎉

build.sbt
```diff
@@ -250,55 +249,27 @@ lazy val docs = (project in file("docs"))
...
-    releaseStepCommandAndRemaining("+publishSigned"),
-    setNextVersion,
-    commitNextVersion,
-    releaseStepCommandAndRemaining("+sonatypeRelease"),
-    pushChanges,
-    releaseStepCommandAndRemaining("publishMicrosite")
+  developers := List(
+    Developer("philwills", "Phil Wills", "", url("https://github.com/philwills")),
+    Developer(
+      "regiskuckaertz",
```
philwills: 🎉

regiskuckaertz:
I had to comment it out because some tests are failing with 2.11 😢 I prefer to investigate separately.

> I seem to remember there was a reason why I originally went for coveralls over scoverage

Let me know if you do, I'm not attached to either ✌️

regiskuckaertz:
@philwills woohoo, the fix turned out to be easier than I thought, I restored the cross build tests.

regiskuckaertz:
I wonder if the package tree should start at `scanamo` instead of `org.scanamo` 🤔

philwills:
I went for org.scanamo as it matches the maven coordinates and fits the java model for trying to avoid clashes.

testkit/src/main/scala/com/gu/scanamo/LocalDynamoDB.scala
```diff
@@ -13,6 +14,7 @@ object LocalDynamoDB {
       .asyncBuilder()
       .withCredentials(new AWSStaticCredentialsProvider(new BasicAWSCredentials("dummy", "credentials")))
       .withEndpointConfiguration(new EndpointConfiguration("http://localhost:8042", ""))
+      .withClientConfiguration(new ClientConfiguration().withClientExecutionTimeout(50000).withRequestTimeout(5000))
```
TBonnin: why do we need this new line?

regiskuckaertz: yes. it was always some alpakka test failing so i increased the patience config there, but this line above cannot hurt anyway.

regiskuckaertz:
@philwills good point! will try this out right now 😄

PR #293, opened 2019-01-09
Added a visualisation of code evolution

gitviscode:
If you like it, feel free to merge it :)


TBonnin:
The video is unavailable to me. :)
Anyway I don't think such visualisation adds much value for scanamo users or developers.

Issue #249, opened 2018-08-27
DynamoFormat doesn't seem to be able to convert Float types


dollschasingmen:
not sure if there is some way to write our own extension for this since `coerceNumber` is private.  It is unexpected that it doesn't handle float.

regiskuckaertz:
You can derive an instance using `DynamoFormat.coercedXmap`. I agree that if we have an instance for `Int`, we should have one for `Float` too.

PR #295, opened 2019-01-10
Add missing floatFormat and floatSetFormat to DynamoFormat

shtukas:
Addressing issue: https://github.com/scanamo/scanamo/issues/249


shtukas:
Hehe :)

Issue #333, opened 2019-01-31
Bumping aws sdk version v2


ratoshniuk:
Hi there!

I am currently working on dynamo db wrapping but I found this library and its API looks pretty good. BUT!
Amazon published new API https://github.com/aws/aws-sdk-java-v2
What about "bumping" to this version? 

I currently working on with that instead of writing yet another wrapper for dynamo

regiskuckaertz:
Hi @ratoshniuk ! yes, v2 has been on my mind for some time. To be honest I did no plan for it before we release v1, which is pending a couple of bug fixes now. But if you have appetite for porting it over, by all means 🤗 That would be very much appreciated 😗

ratoshniuk:
@regiskuckaertz cool! stay tuned with my PR !

ratoshniuk:
Hi there! I made a migration **EXCEPT** alpakka. 

They currently have an open issue for migration sdk (I am also referenced for that) and I don't actually known how to work with alpaca interop.

We can make wrappers between v1 and v2 api. Do you have any ideas?

ratoshniuk:
Anyway, can somebody check my PR and request changes if needs? Thank you!

regiskuckaertz:
Thank you for this. I'll have a look asap.

regiskuckaertz:
Duplicate of #287

Issue #290, opened 2019-01-03
Get operation can throw when misused.


mac01021:
This may or may not be a bug.  I'll leave that up to the authors of the library.

I have the impression that `get`, `query`, and other operations are intended never to throw an exception (but instead return `Left(someError)`.  However, if you call `get('myKey -> "myVal")` when `'myKey` is not the primary key,  then `exec` will throw an exception from within the AWS dynamodb client.   

In my case, the issue was that `myKey` was only the hashing key and I had neglected to provide a value for the sorting key.  Clearly user error, but also still a thrown exception.

regiskuckaertz:
You're right, this shouldn't happen.

mac01021:
Out of curiosity, what's the reason for `query` returning `List[Either[Error, Record]]` rather than `Either[Error, List[Record]]`?

regiskuckaertz:
See https://github.com/scanamo/scanamo/issues/174

We'll be working out a refactor of that soon.

regiskuckaertz:
Technically, the semantics of `List[Either[]]` is more accurate since each error can occur at the record level. We don't want the whole result to fail if only one record fail, though we don't want to lose that information either. So, I'm not sure `Either[List[]]` is the right choice here. We need a `ScanamoGetResult[A]`.

PR #338, opened 2019-01-31
Feature/aws v2

ratoshniuk:


ratoshniuk:
Fixed #333

ratoshniuk:
Current work with Alpakka is absolutely inconvenient because alpakka uses v1. So i decided to make some wrappers from legacy and back. But it seems to be overhead, and I dunno how to make temporary workaround kinda neat

ratoshniuk:
Also! I make an update  for zio and added docker-compose (i will remove them from this PR) and maybe it also would be OK to rebase master after https://github.com/scanamo/scanamo/pull/329

regiskuckaertz:
Thanks for your great work on this! This ticket is going to be very tricky. Lots of apps are still relying on v1, we don't want to sneakily introduce the v2 sdk there. That would probably create issues at the end of the world anyway. I will review but I'm of the opinion that `ScanamoOps` should be parameterised by the model datatypes, i.e.:
```scala
abstract class ScanamoOps[PutItemResult, GetItemResult, ..., A]
```
and then we could add a `org.scanamo.ops.impl` package where v1 and v2 specialisations would like:
```
sealed abstract class ScanamoOpsV1[A] extends ScanamoOps[PutItemResult, GetItemResult, ...]
```
and 
```
import software.amazon.awssdk.services.dynamodb.model._
sealed abstract class ScanamoOpsV1[A] extends ScanamoOps[PutItemResponse, GetItemResponse, ...]
```

ratoshniuk:
@ratoshniuk Yeah, that makes sense. Ok, I will do that.

TBonnin:
@regiskuckaertz what about abstracting AWS dynamodb types rather than exposing them to the lib end user. Of course that would be a massive change but we already about to do it. See your comment at https://github.com/scanamo/scanamo/issues/290 ;)

ratoshniuk:
@TBonnin I am currently made an abstraction over AttributeValue. That was a start point

ratoshniuk:
@TBonnin @regiskuckaertz It is really good idea to make an abstraction over amazon. it helps to test all api even without real environment

regiskuckaertz:
Yes I think this is the direction I'd like to move in, ultimately. We could then have unit tests to validate the properties of the library and integration tests using v1 and v2 sdks.

@TBonnin The problem is that we have to deal with two different concrete implementations of the dynamodb model, adding data on top of them will make them invisible to the outside world, but I think the challenge here is also to make the core of the library oblivious to them as well.

ratoshniuk:
@regiskuckaertz i almost finish getting rid of aws deps from `scanamo` and `formats`as well. stay tuned

ratoshniuk:
@TBonnin @regiskuckaertz  Hey guys, it seems I confused and need some help

https://github.com/ratoshniuk/scanamo/blob/289141a74b67daf5d742e18b61973381fb067329/formats/src/main/scala/com/gu/scanamo/DynamoFormat.scala#L530

it can't find any implicit implementations. It seems that

```scala
DynamoFormat[T, V] != DynamoFormatV1[T]

type V = ???
but DynamoFormatV1[T] = DynamoFormat[T, V]
```

ratoshniuk:
As I understand it uses simulacrum to provide companion object with apply. But I didn’t manage to make it works when i put another type arg for DynamoFormat. It is an abstraction over AttributeValue

regiskuckaertz:
Hmm, you can try making an existential type instead:
```scala
trait DynamoFormat[T] { type V }
trait DynamoFormatV1[T] extends DynamoFormat[T] { type V = AttributeValue }
```

ratoshniuk:
@regiskuckaertz V is a typeclass, I can not do that

ratoshniuk:
@regiskuckaertz Please, could you jump into this one and try out my approach?
 #345

It became a completely non-obvious problem for me and I think I can't perform proper abstraction.
And I think it's too silly and redundant as I did.

regiskuckaertz:
Major rework in the pipeline that will significantly simplify this, so closing for now.

Would be thrilled to resume when that is ready. Keep jour eyes peeled

PR #361, opened 2019-02-21
Ignore updates for the dynamo db sdk

regiskuckaertz:
I think they've proven enough to be a major pain in the a**


regiskuckaertz:
@TBonnin Let's write that! That would be a super useful project for us and many organisations ❤️

PR #367, opened 2019-03-14
Write empty strings as null

### LINKED ISSUE #366: One or more parameter values were invalid: An AttributeValue may not contain an empty string
I'm experiencing the following error when one of my fields is an empty string:

```
One or more parameter values were invalid: An AttributeValue may not contain an empty string (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException;
```

I was very surprised to discover that DynamoDb doesn't support empty strings.
Is it supposed to work out of the box with Scanamo?
Is there a way to configure Scanamo/provide a string format to work around this issue?

regiskuckaertz:
Fixes #366 

That DynamoDB won't accept empty strings is a long-standing pain in the ass. As of now, we've been encoding strings without taking care of it, but the result is this behaviour bubbles up into user-land code where people must come up with handling mechanisms. It seems we can take care of this ourselves, by encoding empty strings as nulls.

To make matters even worse, when marshalling an object, if a key has a null value, it will be stripped off in DynamoDB. As a result, we must make sure that missing properties of type `String` are properly decoded into the empty string: this is what the `DynanoFormat[String]#default` value is for.

### DIFF
--- a/formats/src/main/scala/org/scanamo/DynamoFormat.scala
+++ b/formats/src/main/scala/org/scanamo/DynamoFormat.scala
@@ -173,8 +173,8 @@ object DynamoFormat extends EnumDynamoFormat {
     * }}}
     */
   implicit val stringFormat: DynamoFormat[String] = new DynamoFormat[String] {
-    val default = Some("")
-    
+    override val default = Some("")
+
     def read(av: AttributeValue): Either[DynamoReadError, String] =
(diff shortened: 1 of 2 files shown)


shtukas:
Proposal makes sense. I can't see any reason right now why it would be a problem. +1

regiskuckaertz:
Thanks 🙏

PR #373, opened 2019-03-29
Update put to handle exceptions from Dynamo.

susiecoleman:
Also removed conditional put.

Following on from https://github.com/scanamo/scanamo/pull/365 experiment with how the error handling could be done in Scanamo in order to get feedback.

UPDATE: This PR is now updated to remove conditional update and conditional delete. This has changed the way that we do error handling and this solution is not ideal. Potentially the project is in a position where we should consider moving away from using a free monad.

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
+++ b/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
@@ -2,7 +2,7 @@ package org.scanamo
 
 import com.amazonaws.services.dynamodbv2.model.{PutRequest, WriteRequest, _}
 import org.scanamo.DynamoResultStream.{QueryResultStream, ScanResultStream}
-import org.scanamo.error.DynamoReadError
+import org.scanamo.error.{DynamoDBException, DynamoReadError, ScanamoError}
 import org.scanamo.ops.ScanamoOps
 import org.scanamo.query._
 import org.scanamo.request._
@@ -16,19 +16,23 @@ object ScanamoFree {
 
   private val batchSize = 25
 
-  def put[T](tableName: String)(item: T)(implicit f: DynamoFormat[T]): ScanamoOps[Option[Either[DynamoReadError, T]]] =
+  def put[T](tableName: String)(item: T)(implicit f: DynamoFormat[T]): ScanamoOps[Option[Either[ScanamoError, T]]] =
     ScanamoOps
       .put(
         ScanamoPutRequest(tableName, f.write(item), None)
       )
-      .map { r =>
-        if (Option(r.getAttributes).exists(_.asScala.nonEmpty)) {
-          Some(
-            f.read(new AttributeValue().withM(r.getAttributes))
-          )
-        } else {
-          None
-        }
+      .map {
+        _.fold(
+          e => Some(Left(DynamoDBException(e))),
+          r =>
+            if (Option(r.getAttributes).exists(_.asScala.nonEmpty)) {
+              Some(
+                f.read(new AttributeValue().withM(r.getAttributes))
+              )
+            } else {
+              None
+            }
+        )
       }
 
   def putAll[T](tableName: String)(items: Set[T])(implicit f: DynamoFormat[T]): ScanamoOps[List[BatchWriteItemResult]] =
--- a/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
+++ b/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
@@ -4,6 +4,7 @@ import com.amazonaws.services.dynamodbv2.model.{PutRequest, WriteRequest, _}
 import org.scanamo.DynamoResultStream.{QueryResultStream, ScanResultStream}
 import org.scanamo.error.{DynamoDBException, DynamoReadError, ScanamoError}
 import org.scanamo.ops.ScanamoOps
+import org.scanamo.ops.ScanamoOpsA.ScanamoResult
 import org.scanamo.query._
 import org.scanamo.request._
 import org.scanamo.update.UpdateExpression
@@ -124,7 +125,7 @@ object ScanamoFree {
       }
       .map(_.flatMap(_.getResponses.get(tableName).asScala.toSet.map(read[T])).toSet)
 
-  def delete(tableName: String)(key: UniqueKey[_]): ScanamoOps[DeleteItemResult] =
+  def delete(tableName: String)(key: UniqueKey[_]): ScanamoOps[ScanamoResult[DeleteItemResult]] =
     ScanamoOps.delete(ScanamoDeleteRequest(tableName, key.asAVMap, None))
 
   def scan[T: DynamoFormat](tableName: String): ScanamoOps[List[Either[DynamoReadError, T]]] =
--- a/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
+++ b/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
@@ -67,19 +67,28 @@ object ScanamoFree {
 
   def get[T](
     tableName: String
-  )(key: UniqueKey[_])(implicit ft: DynamoFormat[T]): ScanamoOps[Option[Either[DynamoReadError, T]]] =
+  )(key: UniqueKey[_])(implicit ft: DynamoFormat[T]): ScanamoOps[Option[Either[ScanamoError, T]]] =
     for {
       res <- ScanamoOps.get(new GetItemRequest().withTableName(tableName).withKey(key.asAVMap.asJava))
-    } yield Option(res.getItem).map(read[T])
+    } yield res.fold(
+        e => Some(Left(DynamoDBException(e))),
+        r =>
+          Option(r.getItem).map(read[T])
+      )
 
   def getWithConsistency[T](
     tableName: String
-  )(key: UniqueKey[_])(implicit ft: DynamoFormat[T]): ScanamoOps[Option[Either[DynamoReadError, T]]] =
+  )(key: UniqueKey[_])(implicit ft: DynamoFormat[T]): ScanamoOps[Option[Either[ScanamoError, T]]] =
     for {
       res <- ScanamoOps.get(
         new GetItemRequest().withTableName(tableName).withKey(key.asAVMap.asJava).withConsistentRead(true)
       )
-    } yield Option(res.getItem).map(read[T])
+    } yield res.fold(
+        e => Some(Left(DynamoDBException(e))),
+        r =>
+          Option(r.getItem).map(read[T])
+      )
+
 
   def getAll[T: DynamoFormat](tableName: String)(keys: UniqueKeys[_]): ScanamoOps[Set[Either[DynamoReadError, T]]] =
     keys.asAVMap
--- a/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
+++ b/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
@@ -142,7 +142,7 @@ object ScanamoFree {
 
   def update[T](tableName: String)(key: UniqueKey[_])(update: UpdateExpression)(
     implicit format: DynamoFormat[T]
-  ): ScanamoOps[Either[DynamoReadError, T]] =
+  ): ScanamoOps[Either[ScanamoError, T]] =
     ScanamoOps
       .update(
         ScanamoUpdateRequest(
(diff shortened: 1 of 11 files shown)


scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
```diff
@@ -16,19 +16,23 @@ object ScanamoFree {
...
-            f.read(new AttributeValue().withM(r.getAttributes))
-          )
-        } else {
-          None
-        }
+      .map {
+        _.fold(
+          e => Some(Left(DynamoDBException(e))),
+          r =>
+            if (Option(r.getAttributes).exists(_.asScala.nonEmpty)) {
```
regiskuckaertz: Wow I just saw that 😮 I think a better expression would be:
```scala
Option(r.getAttributes)
  .filter(!_.isEmpty)
  // ^ no need to do the conversion to a Scala Map here
  .map(attrs => f.read(new AttributeValue().withM(attrs)))
```

scanamo/src/main/scala/org/scanamo/query/ConditionExpression.scala
```diff
@@ -13,12 +13,6 @@ case class ConditionalOperation[V, T](tableName: String, t: T)(
   implicit state: ConditionExpression[T],
   format: DynamoFormat[V]
 ) {
-  def put(item: V): ScanamoOps[Either[ConditionalCheckFailedException, PutItemResult]] = {
```
regiskuckaertz: If you get rid of this, it won't be possible to do conditional puts anymore. I think this method should stay, but the return type changed according to your changes:
```
def put(item: V): ScanamoOps[ScanamoResult[PutItemResult]] = ScanamoOps.put(
  ScanamoPutRequest(tableName, format.write(item), Some(state.apply(t)(None))
)
```

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
```diff
@@ -1,12 +1,14 @@
...
 import com.amazonaws.services.dynamodbv2.model._
+import org.scanamo.ops.ScanamoOpsA.ScanamoResult
 import org.scanamo.request._
 
 sealed trait ScanamoOpsA[A] extends Product with Serializable
-final case class Put(req: ScanamoPutRequest) extends ScanamoOpsA[PutItemResult]
-final case class ConditionalPut(req: ScanamoPutRequest)
-    extends ScanamoOpsA[Either[ConditionalCheckFailedException, PutItemResult]]
+object ScanamoOpsA {
+  type ScanamoResult[A] = Either[AmazonDynamoDBException, A]
```
regiskuckaertz: I'm not so sure about this type anymore. Right now Scanamo is a glorified `A => F[B]` and so the way to surface errors is either:
- `A => F[Either[E, B]]`
- `A => F[E, B]`
- `A => F[B]` (i.e. bake error in the semantics of `F`)

The latter option, which we use at the moment, is the easy one. With the caveat that it throws exceptions when using the synchronous client. I think your solution (the first choice, which was already used for conditional puts) is a good step forward because we want to advertise the kind of errors that can happen during an effect. In the future though, I'm inclined to say the second option is the best, because error-handling is usually baked into effect monads and it will be optimised for that (e.g. in zio there is first-class support for fully typed errors, whereas the current cats IO will have to become `EitherT[IO, ?]`).

susiecoleman: @regiskuckaertz I have gone with the first option for now but it could move towards the second one in the future.

susiecoleman:
@regiskuckaertz It would be great if you could take another look at this. It's still a work in progress. I have changed it so that conditional puts are still possible. I've also changed it so that DynamoDB exceptions are caught when a get is performed

scanamo/src/main/scala/org/scanamo/query/ConditionExpression.scala
```diff
@@ -38,11 +40,11 @@ case class ConditionalOperation[V, T](tableName: String, t: T)(
       None
     )
     ScanamoOps
-      .conditionalUpdate(unconditionalRequest.copy(condition = Some(state.apply(t)(unconditionalRequest.condition))))
+      .update(unconditionalRequest.copy(condition = Some(state.apply(t)(unconditionalRequest.condition))))
       .map(
         either =>
           either
-            .leftMap[ScanamoError](ConditionNotMet(_))
+            .leftMap[ScanamoError](DynamoDBException(_))
```
susiecoleman: @regiskuckaertz & @TBonnin It would be good to get some feedback on this specific change. Switching from `conditionalUpdate` to `update` has made the type of the Left in the Either more general. Does it matter that there is not more granularity in the error?

regiskuckaertz: Hmm, this won't be that easy. We don't want condition checks to break the whole computation, but we cannot say that for other errors. We may want to add a new constructor to `ScanamoOpsA`:
```scala
final case class Fail(t: Throwable) extends ScanamoOpsA[Nothing]
```
and make `ScanamoOps` and instance of `MonadError`:
```scala
implicit val monadErrorScanamoOps = new MonadError[ScanamoOps, Throwable] {
  def raiseError[A](e: Throwable) = liftF[ScanamoOpsA, A](Fail(e))

  def handleErrorWith[A](fa: ScanamoOps[A])(f: Throwable => ScanamoOps[A]) =
    fa resume match {
      case Left(Fail(t)) => f(t)
      case Left(sfa)     => liftF(sfa)
      case Right(a)      => a.pure
    }
}
```
then we can put errors back on the stack for someone to deal with:
```scala
ScanamoOps.update(...)
  .flatMap {
    case Left(t) if t.isInstanceOf[ConditionCheckFailedException] => Left(ConditionNotMet(t)).pure
    case Left(t) => raiseError(t)
    case Right(r) => Right(r).pure
  }
```

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
```diff
@@ -1,48 +1,80 @@
...
-final case class ConditionalDelete(req: ScanamoDeleteRequest)
-    extends ScanamoOpsA[Either[ConditionalCheckFailedException, DeleteItemResult]]
-final case class Scan(req: ScanamoScanRequest) extends ScanamoOpsA[ScanResult]
-final case class Query(req: ScanamoQueryRequest) extends ScanamoOpsA[QueryResult]
-final case class BatchWrite(req: BatchWriteItemRequest) extends ScanamoOpsA[BatchWriteItemResult]
-final case class BatchGet(req: BatchGetItemRequest) extends ScanamoOpsA[BatchGetItemResult]
-final case class Update(req: ScanamoUpdateRequest) extends ScanamoOpsA[UpdateItemResult]
-final case class ConditionalUpdate(req: ScanamoUpdateRequest)
-    extends ScanamoOpsA[Either[ConditionalCheckFailedException, UpdateItemResult]]
+sealed trait ScanamoOpsA[+A] extends Product with Serializable
```
regiskuckaertz: I'm surprised about the variance annotation. Is it because of the `Either` in some ops?

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
```diff
@@ -1,48 +1,80 @@
...
-    liftF[ScanamoOpsA, UpdateItemResult](Update(req))
-  def conditionalUpdate(
+    liftF[ScanamoOpsA, BatchGetItemResult](BatchGet(req, identity))
+  def update(
     req: ScanamoUpdateRequest
-  ): ScanamoOps[Either[ConditionalCheckFailedException, UpdateItemResult]] =
-    liftF[ScanamoOpsA, Either[ConditionalCheckFailedException, UpdateItemResult]](ConditionalUpdate(req))
+  ): ScanamoOps[ScanamoResult[UpdateItemResult]] =
+    liftF[ScanamoOpsA, ScanamoResult[UpdateItemResult]](Update(req, identity))
+//TODO: Does there need to be a fail lift?
```
regiskuckaertz: Yes.

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
```diff
@@ -1,48 +1,80 @@
...
+    }
+
+    override def flatMap[A, B](fa: ScanamoOps[A])(f: A => ScanamoOps[B]): ScanamoOps[B] = fa.flatMap(f)
+    override def tailRecM[A, B](a: A)(f: A => ScanamoOps[Either[A, B]]): ScanamoOps[B] =
+      f(a).flatMap(_ match {
+        case Left(a1) => tailRecM(a1)(f)
+        case Right(b) => pure(b)
+
+      })
+    override def pure[A](a: A): ScanamoOps[A] = Monad[ScanamoOps].pure(a)
```
regiskuckaertz: You can skip the implicit resolution here: `Free.pure(a)`

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
```diff
@@ -1,48 +1,80 @@
...
+    }
+  }
+
+
+  implicit val monadErrorScanamoOps = new MonadError[ScanamoOps, Throwable] {
+    def raiseError[A](e: Throwable): ScanamoOps[A] = liftF[ScanamoOpsA, A](Fail(e))
+
+    override def handleErrorWith[A](fa: ScanamoOps[A])(f: Throwable => ScanamoOps[A]): ScanamoOps[A] =
+      fa.resume match {
+        case Left(Fail(t)) => f(t)
```
regiskuckaertz: 😍

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
```diff
@@ -1,48 +1,80 @@
...
+        case Left(_)  => f(new Throwable("Unexpected exception"))
```
regiskuckaertz: With resume, you either get another computation (`Left`), or you get the final result (`Right`). So here since we have a `Left(op)` where `op : ScanamoOpsA[ScanamoOps[A]]` and we need to return a `ScanamoOps[A]`, what you have to do is essentially put it back on the stack. You can do that with [`roll`](https://typelevel.org/cats/api/cats/free/Free$.html#roll[F[_],A](value:F[cats.free.Free[F,A]]):cats.free.Free[F,A]):
```scala
case Left(op) => Free.roll(op)
```

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
regiskuckaertz: This is actually going to be useful in `ConditionExpression`. Let's imagine this function is called `fail`.

scanamo/src/main/scala/org/scanamo/query/ConditionExpression.scala
```diff
@@ -38,12 +41,15 @@ case class ConditionalOperation[V, T](tableName: String, t: T)(
...
-      .map(
-        either =>
-          either
-            .leftMap[ScanamoError](ConditionNotMet(_))
-            .flatMap(r => format.read(new AttributeValue().withM(r.getAttributes)))
+      .update(unconditionalRequest.copy(condition = Some(state.apply(t)(unconditionalRequest.condition))))
+      .flatMap(
+        _.fold(
+          _ match {
+            case t: ConditionalCheckFailedException => Monad[ScanamoOps].pure(Left(ConditionNotMet(t)))
```
regiskuckaertz: `Free.pure(Left(...))`

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
regiskuckaertz: Scratch that, I was wrong. But yeah, please add the `fail` combinator anyway 🤗

regiskuckaertz:
Thanks for your work on this!! Let's fix these few comments and those conflicts and see how the tests go after that.

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
susiecoleman: It was because of the `Nothing` in the `Fail` case

scanamo/src/main/scala/org/scanamo/ops/ScanamoOpsA.scala
susiecoleman: Added

regiskuckaertz:
@susiecoleman I was off this week and previous week was 🤯. Can we find some time to pair on this one? It may be worth starting from a clean master and a clean history, open a separate PR.

regiskuckaertz:
Closing. Feel free to reopen anytime you have bandwidth to work on this 🤗

PR #400, opened 2019-05-08
Manual derivation with DynamoValue

### LINKED ISSUE #336: Convenience methods for manual derivation
Peeling off the layers of an `AttributeValue` is not necessarily the best thing in life 😁 Nor is it "functional" in any way. We should aim to make writing custom `DynamoFormat` instances a breeze.

regiskuckaertz:
Closes #336 

We need a pure abstraction with a well-defined API to help people easily write manual derivations for their own ADTs. This is `DynamoValue`, an abstract data type isomorphic to `AttributeValue`. It fully represents the whole set of valid attribute values, down to monomorphic specialised arrays. Objects and arrays are represented respectively by `DynamoObject` and `DynamoArray`.

Adding these types means that `DynamoFormat` must be changed like this:
```
trait DynamoFormat[T] {
  def read(av: DynamoValue): Either[DynamoReadError, T]
  def write(t: T): DynamoValue
}
```

And as it turns out, this has impact all over the library 😅 There's still some work that needs doing but this is ready for review. I will cleanup and improve the API, add some property tests and define typeclass instances (e.g. `DynamoObject` has many).

The nice thing about that is we can push the conversion to `AttributeValue` until the very last moment; this will make it easier to support both v1 and v2 of the AWS SDK.

## Summary of changes

- [`DynamoValue`](https://github.com/scanamo/scanamo/pull/400/files#diff-d84ea639301de24601ffe448fea7a8dc) is an abstract representation for `AttributeValue`s respecting the following laws:

  - ∀ a : AttributeValue. DynamoValue.fromAttributeValue(a).toAttributeValue = a
  - ∀ f : AttributeValue → AttributeValue, ∃ g : DynamoValue → DynamoValue s.t. ∀ d : DynamoValue. f(d.toAttributeValue) = g(d).toAttributeValue

- [`DynamoObject`](https://github.com/scanamo/scanamo/pull/400/files#diff-6909f7b3d5dba5d438fa62da7a7512b5) is a wrapper over maps of `DynamoValue`s and is the goto data type for representing product and coproduct types. For example the format for a case class:
```scala
case class Farmer(name: String, age: Int)
```
can be derived as follows:
```
new DynamoFormat[Farmer] {
  def read(dv: DynamoValue) = dv.asObject { o =>
    for {
      name <- o := "name" as[String]
      age  <- o := "age"  as[Int]
    } yield Farmer(name, age)
  
  def write(f: Farmer) = DynamoValue("name" -> f.name, "age" -> f.age)
}
```

~~(I still have to add the `as` combinator but that's the goal)~~ (done)

- [`DynamoArray`](https://github.com/scanamo/scanamo/pull/400/files#diff-b33d34a3b7fc3f971ed00fa574542cad) handles the four flavours of arrays supported by DynamoDB: numbers, strings, bytes and attribute values.

- I have simplified the shapeless derivation to work on `DynamoObject`s directly, as this will be the case 99% of the time. People who want a different encoding will write their instance by hand.

- the type `EvaluationKey` that I hated to introduce is now replace by `DynamoObject` since it can easily take in a java `Map` ... or produce one, avoiding extra conversions when possible

- **Breaking change**: queries/scans in descending order are now done with
```scala
table.descending.query(???)
```
rather than
```scala
table.query(???.descending)
```
because it made the creation of Dynamo requests much easier (it's also in line with all other query modifiers like `index`, `limit`, `from`, `consistently`, etc.

- [refactored `QueryableKeyCondition`](https://github.com/scanamo/scanamo/pull/400/files#diff-d20558a45e02d29ea28609685d5c43fbL15) to produce a `RequestCondition` instead of a `QueryRequest => QueryRequest`. The reason is that is in doing so there is only one point where the query request is created and filled with all the parameters. Made things easier to follow.

- the changes in [`UpdateExpression`](https://github.com/scanamo/scanamo/pull/400/files#diff-ddc8d8c78b3308704135370f39202182) are mostly reordering the code for clarity as it was super hard for me to follow.

After reading the code, there is still much simplification that could be done. I don't expect to catch all of them, but overall just improving the API of the new primitives should already go a long way.


formats/src/main/scala/org/scanamo/DerivedDynamoFormat.scala
```diff
@@ -1,31 +1,29 @@
...
 import org.scanamo.error._
 import org.scanamo.export.Exported
 import shapeless._
 import shapeless.labelled._
 
-import collection.JavaConverters._
-
 trait DerivedDynamoFormat {
-  type ValidatedPropertiesError[T] = Validated[InvalidPropertiesError, T]
+  type ValidatedPropertiesError[T] = ValidatedNel[(String, DynamoReadError), T]
```
TBonnin: Would it make sense to have a type alias `fieldname` rather than String, just to document what the first element of the tuple actually is?

formats/src/main/scala/org/scanamo/DerivedDynamoFormat.scala
```diff
@@ -35,75 +33,74 @@ trait DerivedDynamoFormat {
...
 
         cats.Apply[ValidatedPropertiesError].map2(head, tail)(_ :: _)
       }
-      def write(t: FieldType[K, V] :: T): AttributeValue = {
+
+      final def write(t: FieldType[K, V] :: T) = {
         val tailValue = tailFormat.value.write(t.tail)
         val av = headFormat.value.write(t.head)
-        if (!(av.isNULL eq null) && av.isNULL)
+        if (av.isNull)
```
TBonnin: You removed the first part of the condition. Was it unnecessary?

regiskuckaertz:
> That's a pretty big PR, touching a lot of code

Yes indeed. I will properly document all the significant changes I did in the description.

> Do we have enough test coverage to ensure those changes are not breaking anything?

Not nearly enough. I will add some tests for the new primitives, but in a future PR I'd like to improve the testing strategy.

> I started having a look but I need to head out now.

Thank you 🙏

formats/src/main/scala/org/scanamo/DerivedDynamoFormat.scala
regiskuckaertz: `av : DynamoValue` which [fixes the behaviour](https://github.com/scanamo/scanamo/pull/400/files#diff-d84ea639301de24601ffe448fea7a8dcR179) of null handling in `AttributeValue`.

regiskuckaertz:
2.11 does not support SAM types, which breaks compilation. I'm now wondering if we should just drop 2.11. Opinion @TBonnin ?

formats/src/main/scala/org/scanamo/DerivedDynamoFormat.scala
```diff
@@ -35,75 +34,74 @@ trait DerivedDynamoFormat {
...
-          av.getM.asScala.get(fieldName).map(headFormat.value.read).orElse(headFormat.value.default.map(Either.right))
+      private final val fieldName = fieldWitness.value.name
 
-        val valueOrError = possibleValue.getOrElse(Either.left[DynamoReadError, V](MissingProperty))
+      final def read(av: DynamoObject) = {
+        val valueOrError: Either[(FieldName, DynamoReadError), V] =
+          av(fieldName).map(headFormat.value.read) orElse
+            headFormat.value.default.map(Either.right) getOrElse
+            Either.left(MissingProperty) leftMap
+            ((fieldName, _))
```
TBonnin: I have never been a huge fan of mixing dot and space notation. Especially in the same function or even the same line

formats/src/main/scala/org/scanamo/DynamoArray.scala
```diff
@@ -0,0 +1,220 @@
...
+          ),
+          concatDynamoValues
+        )
+      )
+    case Pure(xs) => Some(xs)
+    case Concat(xs0, ys0) =>
+      for {
+        xs <- xs0.asArray
+        ys <- ys0.asArray
+      } yield xs ++ ys
```
TBonnin: we use for comprehension a lot in place where a good old applicative would be enough. ie: `(xs0.asArray, ys0.asArray).mapN(_ ++ _)`
What do you think?

TBonnin:
@regiskuckaertz regarding dropping support for 2.11: Is there a way to know how many projects using Scanamo run on scala 2.11. Are maven download stats available for instance?

formats/src/main/scala/org/scanamo/DynamoObject.scala
```diff
@@ -0,0 +1,187 @@
...
+  final def toMap[V](implicit D: DynamoFormat[V]): Either[DynamoReadError, Map[String, V]] = self match {
+    case Empty => Right(Map.empty)
+    case Strict(xs) =>
+      xs.entrySet.stream.reduce[Either[DynamoReadError, Map[String, V]]](
+        Right(Map.empty),
+        jfun2[Either[DynamoReadError, Map[String, V]], JMap.Entry[String, AttributeValue], Either[
+          DynamoReadError,
+          Map[String, V]
+        ]]((xs0, e) => for { xs <- xs0; x <- D.read(e.getValue) } yield xs + (e.getKey -> x)),
+        jbiop[Either[DynamoReadError, Map[String, V]]]((xs0, ys0) => for { xs <- xs0; ys <- ys0 } yield xs ++ ys)
```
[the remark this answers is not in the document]
regiskuckaertz: No it isn't. I did that because 2.11 doesn't support SAM types but in fact it does if you add `-Xexperimental` to the compiler flags. Next update will remove all that boilerplate.

formats/src/main/scala/org/scanamo/DynamoArray.scala
```diff
@@ -0,0 +1,220 @@
...
+  import DynamoArray._
+
+  def apply(i: Int): Option[DynamoValue] = self match {
+    case Empty       => None
+    case Strict(xs)  => Option(xs.get(i)).map(DynamoValue.fromAttributeValue)
+    case StrictS(xs) => Option(xs.get(i)).map(DynamoValue.string)
+    case StrictN(xs) => Option(xs.get(i)).map(DynamoValue.unsafeNumber)
+    case StrictB(xs) => Option(xs.get(i)).map(DynamoValue.byteBuffer)
+    case Pure(xs)    => xs.drop(i).headOption
+    case PureS(xs)   => xs.drop(1).headOption.map(DynamoValue.string)
```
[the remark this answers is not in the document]
regiskuckaertz: 🤦‍♂ good catch!

formats/src/main/scala/org/scanamo/DynamoArray.scala
regiskuckaertz: Great idea!

regiskuckaertz:
No idea, let's investigate after this PR

regiskuckaertz:
We still have loads at the G, so probably not an option for now.

regiskuckaertz:
I'm going to merge this and release a tag version.

> Are we confident the unit tests have good enough coverage?

No, but the tests in the `Table` and `SecondaryIndex` scaladoc are more comprehensive 🤞

Issue #397, opened 2019-05-02
InternalServerErrorException: Internal server Error


AyushTiwary:
I am using ScanamoAlpakka in one of my projects, and we are facing a problem i.e. any random request is failing with Internal Server Error. The occurrence of facing this exception is once or twice a month.

I am also using DynamoDB in my other projects too where we are not using Scanamo there everything works fine but this is not happening in the case of Scanamo. 

One of the example of such exception is as follow:

com.amazonaws.services.dynamodbv2.model.InternalServerErrorException: Internal server error (Service: dynamodb; Status Code: 500; Error Code: InternalServerError; Request ID: randonStringForRequestID)

regiskuckaertz:
Hi @AyushTiwary I'm sorry to hear that, it's not good. The message isn't very helpful 🤔 Do you think you could provide an example of a query that caused the error to happen?

AyushTiwary:
Hey @regiskuckaertz, It's not about the query. As every time it occurs randomly on any query. Even in the normal CRUD operations.
These queries even working fine in the rest of the time. Just failed with internal server error randomly on any request.

I have a few questions regarding the Scanamo implementation for DynamoDB.
Are you using one of AWS SDKs?
If Not, then Is scanamo support self retry mechanism that is by default available if we directly use the AWS SDK for DynamoDB?
As this problem is very common in case of DynamoDB. That for any millisecond, InternalServer errors are expected every once in a while, SDK clients will do automatic retries until the request goes through, to your application requests will just look successful.

regiskuckaertz:
We are using the v1 SDK at the moment, yes. The retry mechanism is set in the client configuration, which you have to supply in order to run a scanamo op via any of the interpreters. Is it possible that the alpakka one does not give you access to it?

AyushTiwary:
Can you please tell me how can we supply those retry configurations to the client?
As I don't find anything related to it in Scanamo documentation.
Yes, I am using ScanamoAlpakka.

I am creating the DynamaDBClient as follow:
val dynamoDBClient : DynamoClient = DynamoClient(DynamoSettings(system))
One of the sample query is as follow:
ScanamoAlpakka.exec(dynamoDBClient.m_client)(UserTable.query(PrimaryKey -> value))


Maybe this will help you to understand my issue.

regiskuckaertz:
[...]

PR #402, opened 2019-05-10
Issue/397 : Added retry mechanism for AlpakkaScanamo Interpretor

AyushTiwary:
This PR is against issue #397. I added the retry mechanism in the case of specific DynamoDBException suggested by AWS DynamoDB [official documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html).


alpakka/src/main/scala/org/scanamo/RetryUtility.scala
```diff
@@ -0,0 +1,43 @@
...
+                          retrySettings: RetrySettings)(implicit executionContext: ExecutionContext): Future[T] =
+    op.recoverWith {
+      case exception @ (_: InternalServerErrorException | _: ItemCollectionSizeLimitExceededException |
+          _: LimitExceededException | _: ProvisionedThroughputExceededException | _: RequestLimitExceededException) =>
+        val retries = retrySettings.retries
+        val initialDelay = retrySettings.initialDelay.toMillis
+        val factor = retrySettings.factor
+        if (retries > 0) {
+          for {
+            _ <- waitForMillis(initialDelay)
```
regiskuckaertz: Same here, let's move that to `RetrySettings`:
```scala
trait RetrySettings {
  def delay: Long
}
```

Now i think `RetrySettings` should be named `RetryPolicy`.

alpakka/src/main/scala/org/scanamo/RetryUtility.scala
```diff
@@ -0,0 +1,43 @@
...
+            newRetrySetting = RetrySettings((initialDelay * factor).millis, factor, retries - 1)
+            response <- retryWithBackOff(op, newRetrySetting)
+          } yield response
+        } else {
+          Future.failed(exception)
+        }
+    }
+
+  private def waitForMillis(millis: Long) = {
```
regiskuckaertz: There is no way to cancel from the outside, which is sad with scala futures 😞 I'll try to think of something

alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
```diff
@@ -12,7 +12,7 @@ import scala.concurrent.{ExecutionContext, Future}
 object ScanamoAlpakka {
   import cats.instances.future._
 
-  def exec[A](client: DynamoClient)(op: ScanamoOps[A])(implicit ec: ExecutionContext): Future[A] =
-    op.foldMap(AlpakkaInterpreter.future(client)(ec))
+  def exec[A](client: DynamoClient)(op: ScanamoOps[A])(implicit ec: ExecutionContext, retrySettings : RetrySettings): Future[A] =
```
regiskuckaertz: I think the retry policy should be provided explicitly by the user, we want them to make an informed choice about that. So let's make it an explicit parameter everywhere.

alpakka/src/main/scala/org/scanamo/RetrySettings.scala
```diff
@@ -0,0 +1,5 @@
+package org.scanamo
+
+import scala.concurrent.duration.{FiniteDuration, _}
+
+case class RetrySettings(initialDelay: FiniteDuration = 20.millis, factor: Int = 2, retries: Int = 10)
```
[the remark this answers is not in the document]
regiskuckaertz: Also we need some predefined constants: `never`, `forever`, `once`.

alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
AyushTiwary: Defining the retry policy as explicitly here it's not the good choice according to me. As we define it explicitly then whenever anybody makes the call has to pass the retry policy which reduces the Scanamo clarity for the queries.

I personally choose it as implicit so that the user just needs to define only once and then use of every query. Even if we make it implicit then also give the user the informed choice about that because the user has to define its policy (of implicit type) in the scope to run the query.

alpakka/src/main/scala/org/scanamo/RetryUtility.scala
AyushTiwary: @regiskuckaertz initialDelay is already coming from RetrySettings.
Can you please elaborate your suggestion?

regiskuckaertz: The waiting time (`initialDelay` or what I call `delay` above) is a function of the particular policy:

- a constant policy will always yield the same amount
- a linear one will yield `n * d`
- an exponential one will yield `n ^ f * d`

Right now you have implemented the second, but we'll add more and I think it's best if the loop that you wrote doesn't concern itself with the details of the calculation. How about this:

```scala
sealed trait RetryPolicy { self =>
  final val done: Boolean = self match {
    case Max(0)    => true
    case And(x, y) => x.done && y.done
    case Or(x, y)  => x.done || y.done
    case _         => false
  }

  final val delay: Long = self match {
    case Constant(d)          => d.toMillis
    case Linear(d, n)         => d.toMillis * n
    case Exponential(d, n, f) => d.toMillis * Math.pow(n, f)
    case And(x, y)            => Math.max(x.delay, y.delay)
    case Or(x, y)             => Math.min(x.delay, y.delay)
    case _                    => 0
  }

  final def update: RetryPolicy = self match {
    case x: Constant    => x
    case x: Linear      => x.copy(n = n + 1)
    case x: Exponential => x.copy(n = n + 1)
    case Max(m)         => Max(m - 1)
    case And(x, y)      => And(x.update, y.update)
    case Or(x, y)       => Or(x.update, y.update)
  }

  final def &&(that: RetryPolicy): RetryPolicy = And(self, that)
  final def ||(that: RetryPolicy): RetryPolicy = Or(self, that)
}

object RetryPolicy {
  private[RetryPolicy] final case class Constant(d: FiniteDuration) extends RetryPolicy
  private[RetryPolicy] final case class Linear(d: FiniteDuration, n: Int) extends RetryPolicy
  private[RetryPolicy] final case class Exponential(d: FiniteDuration, n: Int, f: Double) extends RetryPolicy
  private[RetryPolicy] final case class Max(m: Int) extends RetryPolicy
  private[RetryPolicy] final case class And(x: RetryPolicy, y: RetryPolicy) extends RetryPolicy
  private[RetryPolicy] final case class Or(x: RetryPolicy, y: RetryPolicy) extends RetryPolicy
}
```

With that we can express many flavours of policies, with the addition perhaps of two constructors: `Always` and `Never`.

alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
regiskuckaertz: You are completely right 👍 In fact, we can say the same about the `client`. Look at the `foldMap`, we create a new natural transformation every time we call `exec`, even though it will always be the same! I think all these parameters should bubble up to `ScanamoAlpakka`:
```scala
class ScanamoAlpakka(
  client: DynamoClient,
  retryPolicy : RetryPolicy = RetryPolicy.default
)(implicit ec: ExecutionContext) extends ScanamoOpsA ~> Future {
  final def exec[A](op: ScanamoOps[A]): Future[A] = apply(op)

  final override def apply[A](ops: ScanamoOpsA[A]) = ...
}
```

I'm happy to do that in a separate PR as it is a wholesome change to all the interpreters, so I'm fine with the retry being an implicit for now.

alpakka/src/main/scala/org/scanamo/RetryUtility.scala
regiskuckaertz: I think we also want a `Capped(m: Long, p: RetryPolicy)` that will use `p.delay` but cap at `m`.

AyushTiwary: @regiskuckaertz If I understand you correctly, As I added the number of retried then there is no need to add always & never cap too.

Moreover, I added the number of policies that you suggest to add in it.
Please take a look at it

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -0,0 +1,45 @@
...
+    case Max(retries)                        => Max(retries - 1)
+    case And(x, y)                           => And(x.update, y.update)
+    case Or(x, y)                            => Or(x.update, y.update)
+  }
+}
+
+object RetryPolicy {
+  final case class Constant(retryDelay: FiniteDuration, numberOfRetries: Int) extends RetryPolicy
+  final case class Linear(retryDelay: FiniteDuration, numberOfRetries: Int, factor: Double) extends RetryPolicy
+  final case class Exponential(retryDelay: FiniteDuration, numberOfRetries: Int, factor: Double) extends RetryPolicy
```
regiskuckaertz: I would drop `numberOfRetries` from the three above so that we can express the same thing with:
```scala
Linear(delay) && Max(numberOfRetries)
```
Now we have nice decoupling and more potential for combinations.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -0,0 +1,45 @@
...
+    case policy @ Constant(_, retries)       => policy.copy(numberOfRetries = retries - 1)
+    case policy @ Linear(_, retries, _)      => policy.copy(numberOfRetries = retries - 1)
+    case policy @ Exponential(_, retries, _) => policy.copy(numberOfRetries = retries - 1)
+    case Max(retries)                        => Max(retries - 1)
+    case And(x, y)                           => And(x.update, y.update)
+    case Or(x, y)                            => Or(x.update, y.update)
+  }
+}
+
+object RetryPolicy {
```
regiskuckaertz: It's not possible to represent "always" right now but "never" we can:
```scala
val Never: RetryPolicy = Max(0)
```

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -0,0 +1,45 @@
+package org.scanamo.ops.retrypolicy
+
+import org.scanamo.ops.retrypolicy.RetryPolicy._
+import scala.concurrent.duration._
+
+sealed trait RetryPolicy { self =>
```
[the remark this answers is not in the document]
regiskuckaertz: Also turn it into a serialisable class:
```scala
sealed abstract class RetryPolicy extends Product with Serializable
```

alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
```diff
@@ -12,7 +19,11 @@ import scala.concurrent.{ExecutionContext, Future}
...
   import cats.instances.future._
 
-  def exec[A](client: DynamoClient)(op: ScanamoOps[A])(implicit ec: ExecutionContext): Future[A] =
-    op.foldMap(AlpakkaInterpreter.future(client)(ec))
-
+  def exec[A](
+    client: DynamoClient
+  )(op: ScanamoOps[A], retrySettings: RetryPolicy = DefaultRetryPolicy.defaultPolicy)(
+    implicit ec: ExecutionContext,
+    scheduler: ScheduledExecutorService = DefaultRetryPolicy.defaultScheduler
```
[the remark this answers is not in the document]
AyushTiwary: Sure, I will add the TODO for another PR.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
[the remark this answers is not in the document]
AyushTiwary: Sure

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
AyushTiwary: I will think over it. Is there any way to achieve it? Thanks for pointing it out.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
AyushTiwary: This all possible just because of your suggestions.

AyushTiwary: Actually, I added numberOfRetries to keep that in mind, the user does not have to think twice for common retry mechanism.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
AyushTiwary: Added the new RetryPolicy for both always and never explicitly.

AyushTiwary:
@regiskuckaertz Thanks for reviewing the PR and also for your awesome suggestion to make the PR more stronger.
I was also thinking about the test cases that how can we integrate the test for retry mechanism. 
I also did with all the changes suggesting while reviewing the PR. Please have a look at it.
Once it will be done, then I will surely try to create the test scenarios in my next PR as soon as possible.

alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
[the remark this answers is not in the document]
mavericksid: @regiskuckaertz and @AyushTiwary tried removing the scheduler from this place, have a look.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -0,0 +1,76 @@
...
+  final def maybeScheduler: Option[ScheduledExecutorService] = self match {
+    case policy @ Constant(_, _)       => Option(policy.scheduler)
+    case policy @ Linear(_, _, _)      => Option(policy.scheduler)
+    case policy @ Exponential(_, _, _) => Option(policy.scheduler)
+    case And(thisPolicy, thatPolicy)   => thisPolicy.maybeScheduler orElse thatPolicy.maybeScheduler
+    case Or(thisPolicy, thatPolicy)    => thisPolicy.maybeScheduler orElse thatPolicy.maybeScheduler
+    case _                             => None
+  }
+
+  final def done: Boolean = self match {
```
regiskuckaertz: I had a flash of insight yesterday. This `done` is false. If you have `And(Max(1), Constant(2.s))` it will keep on going infinitely, even after the max becomes `Max(0)`. What we need instead is
```scala
def continue = self match {
  case Max(0)    => false
  case And(x, y) => x.continue && y.continue
  case Or(x, y)  => x.continue || y.continue
  case _         => true
}
```

Remember to remove `numberOfRetries` from `Linear`, `Constant` and `Exponential`.

regiskuckaertz:
Looking great! Not convinced that shoving the scheduler inside the data is ideal, but let's give this a try. Thanks again for all your hard work and for committing to adding tests in a follow-up ❤️ The only remaining thing would be to turn `done` into `continue` and then:
```scala
if (policy.continue)
  for { ... } yield ...
else
  Future.failed(exception)
```

Have you tried this with your project?

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
regiskuckaertz: Oh, wait, actually no it works with `done` since the test is negated! Sorry about that. Let's keep it as it is then, with only `numberOfRetries` to go away.

mavericksid: @regiskuckaertz 

> I had a flash of insight yesterday. This done is false. If you have And(Max(1), Constant(2.s)) it will keep on going infinitely, even after the max becomes Max(0).

I guess it'll not go on infinitely based on the initial dry run without writing test cases. `And(Max(1), Constant(2.s))` will finally get to `And(Max(0), Constant(2.s))` because of `And(thisPolicy.update, thatPolicy.update)` which will update the Max(1) to Max(0) and then 
`case And(thisPolicy, thatPolicy)        => thisPolicy.done && thatPolicy.done`
will give true and true respectively which will end the loop. I hope I did not miss anything :)

> Remember to remove numberOfRetries from Linear, Constant and Exponential.
If retries with delay as 0 is not a requirement, which is what happens in case of using Max, then Max doesn't look like a good choice and numberOfRetries can stay in Linear, Constant and Exponential. What do you think?

regiskuckaertz: I'd like to keep it as the documentation states that for internal server errors, it is safe to retry without waiting.

mavericksid:
> Looking great! Not convinced that shoving the scheduler inside the data is ideal, but let's give this a try. Thanks again for all your hard work and for committing to adding tests in a follow-up ❤️

Yes, this is not particularly ideal but if you feel it is okay for now then it should be okay.

> Have you tried this with your project?

Not yet, all scenarios are not tried on the project but a few will need to try all of them.

AyushTiwary:
Hey @regiskuckaertz, I manually tested the retry mechanism for many scenarios. It's working as per our expectations.

regiskuckaertz:
Nice! Just run scalafmt and the build should pass.

AyushTiwary:
@regiskuckaertz actually I tried many times to format the retry utilities but it did not. Even in formattimng, it shows everything is formatted. So, I am unable to debug what is the problem with scalafmt in that class.

regiskuckaertz:
It's probably the spec, then `test:scalafmt`?

regiskuckaertz:
Hmm, this is weird:
```
/home/travis/build/scanamo/scanamo/alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala isn't formatted properly!
```

AyushTiwary:
Hey @regiskuckaertz build is passing now

AyushTiwary:
Hey, @regiskuckaertz Thanks for merging the PR. :heart_eyes:
Also just wanted to understand how does the release work?

regiskuckaertz:
Every change to master is published in the sonatype snapshot repo, this one is `0.10+63-1fc3ceb1-SNAPSHOT`. I can release a tag version but would prefer to wait until a few PRs have gone in.

AyushTiwary:
I am trying to use the `0.10+63-1fc3ceb1-SNAPSHOT` but did find any dependency for it.

AyushTiwary:
@regiskuckaertz Is there any private resolver for the snapshot versions?

regiskuckaertz:
you just need to add [another resolver](https://www.scala-sbt.org/1.x/docs/Resolvers.html) in your SBT config

AyushTiwary:
Thanks for the response but there is no update snapshot jar for alpakka dependency.
https://oss.sonatype.org/content/repositories/snapshots/com/gu/

regiskuckaertz:
Ah yeah, we move the repo to a new organisation:
https://oss.sonatype.org/content/repositories/snapshots/org/scanamo/

AyushTiwary:
Thanks, it works

PR #405, opened 2019-05-17
Upgrade to alpakka v1.0.1

regiskuckaertz:
Brings the alpakka interpreter up to date with the latest version.

**Breaking change**: looking at [the documentation](https://doc.akka.io/docs/alpakka/current/dynamodb.html), it seems more idiomatic to use `Source[A, NotUsed]` instead of `Future[A]` as the output type.

@AyushTiwary if you could provide feedback that would be appreciated! Now I just need to figure out how to adapt the retry mechanism you just added.


regiskuckaertz:
Alright, this is ready for a final review now, I'll do the scalafmt stuff in a separate PR.

AyushTiwary:
@regiskuckaertz I just saw the notification. I will surely take a look at it on priority & will be notifying you accordingly

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -39,39 +25,30 @@ sealed abstract class RetryPolicy extends Product with Serializable { self =>
...
+    case policy: Exponential         => policy
     case Max(retries)                => Max(retries - 1)
     case And(thisPolicy, thatPolicy) => And(thisPolicy.update, thatPolicy.update)
     case Or(thisPolicy, thatPolicy)  => Or(thisPolicy.update, thatPolicy.update)
     case retryPolicy                 => retryPolicy
   }
 
+  final val retries: Int = self match {
+    case Max(retries) => retries
+    case And(x, y)    => Math.min(x.retries, y.retries)
```
AyushTiwary: This should be case And(x, y)    => Math.max(x.retries, y.retries), if we want to complete the retries for both

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -39,39 +25,30 @@ sealed abstract class RetryPolicy extends Product with Serializable { self =>
   }
 
   final def update: RetryPolicy = self match {
-    case policy @ Constant(_, numberOfRetries)  => policy.copy(numberOfRetries = numberOfRetries - 1)(policy.scheduler)
-    case policy @ Linear(_, numberOfRetries, _) => policy.copy(numberOfRetries = numberOfRetries - 1)(policy.scheduler)
-    case policy @ Exponential(_, numberOfRetries, _) =>
-      policy.copy(numberOfRetries = numberOfRetries - 1)(policy.scheduler)
+    case policy: Constant            => policy
```
AyushTiwary: If we are not changing anything in the policy then these cases will automatically cover by the last case (by default)

AyushTiwary: Therefore, no need to add retry policy like Constant, Linear and Exponential if it's sound okay to you

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
regiskuckaertz: Good catch!

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
regiskuckaertz: Yes 👍

regiskuckaertz:
This is ready for a final review. After it goes live, I'll publish a new milestone.

AyushTiwary:
Just to understand the terminology. What do you mean by go live? Are you guys to release a tag version or just you are going to merge it into master?

regiskuckaertz:
That's my bad english 😆 I'll merge to master then release a tag version. Just getting the scalafmt one for good measure.

regiskuckaertz:
@AyushTiwary Thank you! BTW, I'd be glad to know how it goes with the latest changes. Also, have a look at [`RestartSource`](https://doc.akka.io/api/akka/current/akka/stream/javadsl/RestartSource$.html) which goes a long way already. I tried to use it but couldn't find a way to short-circuit the retry in case of an error where retry shouldn't occur.

AyushTiwary:
Okay, Will take a look at it.
Just to inform you, I am already doing some sort of testing on the snapshot version to verify the new change. That's why eagerly waiting for the new tag version. So that I can fix my test results to use in the project.

AyushTiwary:
Yeah, I saw the scalafmt one. It also looks ready to me.

AyushTiwary:
> @AyushTiwary Thank you! BTW, I'd be glad to know how it goes with the latest changes. Also, have a look at [`RestartSource`](https://doc.akka.io/api/akka/current/akka/stream/javadsl/RestartSource$.html) which goes a long way already. I tried to use it but couldn't find a way to short-circuit the retry in case of an error where retry shouldn't occur.

A RestartSource wraps a Source that gets restarted when it completes or fails. So, I am not sure that it's the solution for us. As we want the retry only at some specific failures.

PR #412, opened 2019-05-25
[WIP] Switch from tut/Jekyll to mdoc

regiskuckaertz:
I couldn't remember how to setup Jekyll and it pained me to see the Ruby setup take so much time during the travis build, so I switched to mdoc instead. It's fully integrated and does all the heavy-lifting for us, which is a huge win.

There's still a few tasks to perform

- [ ] make sure content is displayed correctly
- [ ] reuse as much of the existing CSS as possible
- [ ] add a favicon
- [ ] add a logo
- [ ] make sure the CNAME is correctly published to the `gh-pages` branch (this is the alias for `www.scanamo.org`)


regiskuckaertz:
I don't have bandwidth to work on that now, let's try to fix the Jekyll stuff first.

PR #419, opened 2019-06-02
Added jitter in Retry-Mechanism for Alpakka Interpreter

AyushTiwary:


alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -14,11 +15,16 @@ sealed abstract class RetryPolicy extends Product with Serializable { self =>
   }
 
   final def delay: Long = self match {
-    case Constant(retryDelay)       => retryDelay.toMillis
-    case Linear(retryDelay, factor) => retryDelay.toMillis * factor.toLong
+    case Constant(retryDelay) =>
```
regiskuckaertz: How about adding a `case class Jitter(min: Double, max: Double, policy: RetryPolicy)` so that you can just add a case:
```scala
case Jitter(min, max, policy) => getDelayWithJitter(min, max, policy.delay)
```

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -40,6 +46,11 @@ sealed abstract class RetryPolicy extends Product with Serializable { self =>
 
   final def &&(that: RetryPolicy): RetryPolicy = And(self, that)
   final def ||(that: RetryPolicy): RetryPolicy = Or(self, that)
+
+  final private def getDelayWithJitter(delayInMillis: Long) = {
+    val maxJitter = Math.ceil(delayInMillis * 0.2).toInt
```
regiskuckaertz: I would instead force the user to supply lower and upper bounds instead of _[0..delay*0.2]_, for three reasons:
- some users may not want any jitter at all
- some users may want to introduce jitter no matter what, say minimum 10ms
- having an upper bound calculated from the delay feels rather arbitrary

regiskuckaertz:
Let me know if you need any more help on this one 😃

regiskuckaertz:
Ping

AyushTiwary:
Hey @regiskuckaertz, extremely apologized for a very delayed response. I will quickly follow up on this PR in coming days. Will start looking into your suggestion.

regiskuckaertz:
No need to apologise for anything, we all have busy lives 🤗 Whenever you're ready.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
Divyanshu564: @regiskuckaertz Pattern matching is required to handle cases like:-

1. Constant:- `retryDelay.toMillis`
2. Linear:- `retryDelay.toMillis * factor.toLong`
3. Exponential:- `delayMillis * Math.pow(delayMillis.toDouble, factor).toLong`

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
Divyanshu564: @regiskuckaertz So as you have suggested we have updated the jitter, So now we have lower and upper bound and graph will line between lower and upper bound for example: -

case1: -
```
delay = 1000ms
lowerBound = 0
upperBound = 0
delayWithJitter = 1000ms
```

case2:-
```
delay = 1000
lowerBound = 100
upperBound = 100
delayWithJitter = random value in range[900, 1100]
```

case3:-
```
delay = 1000
lowerBound = 0
upperBound = 100
delayWithJitter = random value in range[1000, 1100]
```

case4:-
```
delay = 1000
lowerBound = 100
upperBound = 100
delayWithJitter = random value in range[900, 1000]
```

AyushTiwary:
@regiskuckaertz have a look whenever you have sometime for it. @Divyanshu564 updated the PR.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -74,20 +87,20 @@ object RetryPolicy {
   def max(n: Int): RetryPolicy = if (n <= 0) never else Max(n)
 
   /** The policy that always retries, waiting `base` between each attempt */
-  def fixed(base: Duration): RetryPolicy =
+  def fixed(base: Duration, min: Long = 0, max: Long = 0): RetryPolicy =
     if (base < Duration.Zero) never
     else if (base == Duration.Zero) always
-    else Constant(base)
+    else Constant(base, min, max)
```
regiskuckaertz: let's make sure that `0 <= min && 0 <= max` in all the constructors.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -28,33 +28,46 @@ sealed abstract class RetryPolicy extends Product with Serializable { self =>
...
+    case Linear(retryDelay, n, min, max)              => Linear(retryDelay, n + 1, min, max)
+    case Exponential(retryDelay, factor, n, min, max) => Exponential(retryDelay, factor, n + 1, min, max)
+    case retryPolicy                                  => retryPolicy
   }
 
   final def &&(that: RetryPolicy): RetryPolicy = And(self, that)
   final def ||(that: RetryPolicy): RetryPolicy = Or(self, that)
+
+  final private def getDelayWithJitter(delayInMillis: Long, lowerBound: Long, upperBound: Long) = {
+    val start = delayInMillis - lowerBound
```
regiskuckaertz: let's make sure that `0 <= start`

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -16,8 +16,8 @@
 
 package org.scanamo.ops.retrypolicy
 
+import java.util.concurrent.ThreadLocalRandom
```
regiskuckaertz: Let's use `SplittableRandom` instead, so we can add some deterministic tests.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
regiskuckaertz: Yes of course! The thing is, jitter would be a property that you add to an existing policy, instead of it being latent, so:
```scala
case class Jitter(random: SplittableRandom, margin: Long, inner: RetryPolicy) extends RetryPolicy

def delay: Long = self match {
  case Constant(x) => x
  ...
  case Jitter(rand, margin, inner) =>
    // recursively calculate the delay
    val delay = inner.delay
    val (lo, hi) = (Math.max(0, delay - margin), delay + margin + 1)
    rand.nextLong(lo, hi)
```

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
regiskuckaertz: Great stuff! What is the advantage for having that level of precision over the "direction" of the jitter?

Divyanshu564:
@regiskuckaertz Could you please take a moment to check the PR again, I have updated the jitter policy.

alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -16,35 +16,40 @@
...
     case _                           => true
   }
 
-  final def delay: Duration = self match {
-    case Constant(retryDelay)               => retryDelay
-    case Linear(retryDelay, n)              => retryDelay * n.toDouble
-    case Exponential(retryDelay, factor, n) => retryDelay * Math.pow(factor, n.toDouble)
-    case And(thisPolicy, thatPolicy)        => thisPolicy.delay.max(thatPolicy.delay)
-    case Or(thisPolicy, thatPolicy)         => thisPolicy.delay.min(thatPolicy.delay)
-    case Never                              => Duration.Inf
```
AyushTiwary: @Divyanshu564 can you please check it either we need it or not as  I am bit  doubt about it?
what you think @regiskuckaertz ?

Divyanshu564: @AyushTiwary `Never` case is required as without this case `Or(thisPolicy: RetryPolicy, thatPolicy: RetryPolicy)` will always calculate the wrong delay due to `Math.min(Math.min(thisPolicy.delay, thatPolicy.delay))`

regiskuckaertz: Yes, durations form a lattice, it is needed otherwise combinations are unsound.

AyushTiwary: true

AyushTiwary:
Thanks @regiskuckaertz for the quick review & for merging the PR.

Let us know if you need our help in any of the other **Scanamo** issues too. It will be really great for us.

PR #420, opened 2019-06-03
Added overloaded exec to return Future

mavericksid:
Hi @regiskuckaertz 

Do you think returning a Future from exec will be idiomatic? If yes, raised a PR for this addition :)

### DIFF
--- a/alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
+++ b/alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
@@ -1,7 +1,6 @@
 package org.scanamo
 
 import akka.NotUsed
-import akka.stream.Materializer
 import akka.stream.alpakka.dynamodb.DynamoClient
 import akka.stream.scaladsl.{ Sink, Source }
 import cats.Monad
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
It can't hurt 😉 I wonder though, why would you be using the alpakka client if you can just grab `ScanamoAsync` and directly get a `Future[_]`?

You'll have to run fmt (perhaps you need to pull the latest master, I changed some rules recently)

mavericksid:
That's a good point but doesn't org.scanamo.ScanamoAsync use 
`com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsync`
which is inherently blocking?

https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-x-released/
`The AWS SDK for Java 2.x utilizes a new, nonblocking SDK architecture built on Netty to support true nonblocking I/O. The 1.11.x version already has asynchronous variants of service clients, but they are a managed thread pool on top of the synchronous clients, so each request still requires its own thread. `

and build.sbt contains
`val awsDynamoDB = "com.amazonaws" % "aws-java-sdk-dynamodb" % "1.11.504"`

Moreover, I guess converting to Future will make codebase composable and might reduce boilerplate. What do you think?

mavericksid:
@regiskuckaertz 

Run `scalafmtAll` and fixed the other issues. The changes are good to be reviewed if you think they are useful 😃

mavericksid:
Great! Glad it helped 😄 

Thank you for quick replies 👍

PR #432, opened 2019-06-11
Allow streaming in constant space

### LINKED ISSUE #181: Possible to stream query results, instead of loading everything into memory?
First: Thanks for a great library!

If I read the source for `DynamoResultStream` correctly, all scan and query results are loaded completely into memory, before it can be processed. We have to deal with rather large datasets and would like to prevent OOM errors. Do you have a suggestion how it would be possible to stream the results? Ideally I would like to operate on a akka `Source[Either[DynamoReadError, T], NotUsed]`

I tried to implement a custom version of `DynamoResultStream`, but weaving this class into the existing DSL is hard / almost impossible, since all classes are private.

Any pointers would help.

### LINKED ISSUE #275: Remove `DynamoResultStream`
Scanamo has logic so that the 1MB limit for the result set size is circumvented, but this applies to all queries/scans regardless of users' intent. It may be problematic in some situations (as in #181 ) and prevents running some computations in constant space.

I propose to get rid of it, or at the very least to make this an opt-in feature, i.e. by default the behaviour of Scanamo strictly follows that of DynamoDB.

What do you think @philwills ?

regiskuckaertz:
Closes #275 
Closes #181 

Adds `[scan|query]To` and `[scan|query]toPaged` that thread in an effect type with monoid properties (`Alternative <: Applicative with MonoidK`). Actually we need a stronger constraint, `MonadPlus`, as this is [required in the interpreter](https://github.com/scanamo/scanamo/pull/432), but that class is missing in cats.

- [x] add default `F ~> G` for the various interpreters
- [x] add more tests
- [x] add documentation
- ~~rewrite `scan` and `query` with `[scan|query]To`~~

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
+++ b/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
@@ -57,9 +57,9 @@ private[scanamo] trait DynamoResultStream[Req, Res] {
         if (results.isEmpty)
           FreeT.liftT(M.empty)
         else {
-          val l1 = limit(req).map(_ - results.length).fold(pageSize)(Math.min(pageSize, _))
+          val l1 = limit(req).fold(pageSize)(l => Math.min(pageSize, l - results.length))
           lastEvaluatedKey(res)
-            .filterNot(_.isEmpty || l1 <= 0)
+            .filterNot(_ => l1 <= 0)
             .foldLeft(FreeT.pure[ScanamoOpsA, M, List[Either[DynamoReadError, T]]](results))(
               (res, k) => res <+> streamMore(withExclusiveStartKey(k)(req), l1)
             )
--- a/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
+++ b/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
@@ -1,9 +1,12 @@
 package org.scanamo
 
-import cats.free.Free
+import cats.Alternative
+import cats.Alternative.ops._
+import cats.free.{ Free, FreeT }
+import cats.syntax.semigroupk._
 import com.amazonaws.services.dynamodbv2.model.{ QueryResult, ScanResult }
 import org.scanamo.error.DynamoReadError
-import org.scanamo.ops.{ ScanamoOps, ScanamoOpsA }
+import org.scanamo.ops.{ ScanamoOps, ScanamoOpsA, ScanamoOpsT }
 import org.scanamo.request.{ ScanamoQueryRequest, ScanamoScanRequest }
 
 private[scanamo] trait DynamoResultStream[Req, Res] {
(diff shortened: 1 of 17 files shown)


TBonnin:
Nitpicking: it might be good to mark such PR as draft until they are ready for review. I always wonder if I should review it or not.

regiskuckaertz:
Yeah totally... how do I do that? 😅

TBonnin:
I don't know if you can move a PR back to draft state. https://github.blog/2019-02-14-introducing-draft-pull-requests/
Next time though

regiskuckaertz:
Ah yeah, that's useful! Will @ you when ready ✌️

regiskuckaertz:
I won't rewrite `scan` as it requires to have a functor for `ScanamoOpsA`, and I'm not going there in this PR.

regiskuckaertz:
@TBonnin ready for review ✌️

TBonnin:
why the name `scan|queryTo`? The `To` make me think there is a target/destination/something to reach. Am I missing something?

regiskuckaertz:
how about `scanM` and `queryM`?

scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
```diff
@@ -43,6 +46,27 @@ private[scanamo] trait DynamoResultStream[Req, Res] {
...
+  )(implicit M: Alternative[M]): ScanamoOpsT[M, List[Either[DynamoReadError, T]]] = {
+    def streamMore(req: Req, l: Int): ScanamoOpsT[M, List[Either[DynamoReadError, T]]] =
+      exec(withLimit(l)(req)).toFreeT[M].flatMap { res =>
+        val results = items(res).map(ScanamoFree.read[T])
+        if (results.isEmpty)
+          FreeT.liftT(M.empty)
+        else {
+          val l1 = limit(req).map(_ - results.length).fold(pageSize)(Math.min(pageSize, _))
+          lastEvaluatedKey(res)
+            .filterNot(_.isEmpty || l1 <= 0)
```
TBonnin: Is the `isEmpty` condition necessary here?

scanamo/src/main/scala/org/scanamo/SecondaryIndex.scala
```diff
@@ -43,6 +44,25 @@ sealed abstract class SecondaryIndex[V] {
...
+  /**
+    * Performs a scan with the ability to introduce effects into the computation. This is
+    * useful for huge tables when you don't want to load the whole of it in memory, but
+    * scan it page by page, with a maximum of `pageSize` items per page..
+    *
+    * @note DynamoDB will only ever return maximum 1MB of data per scan, so `pageSize` is an
+    * upper bound.
+    */
+  def scanToPaged[M[_]: Alternative](pageSize: Int): ScanamoOpsT[M, List[Either[DynamoReadError, V]]]
+
```
TBonnin: Would `paginated` better than `paged`? I am not a native english speaker so feel free to ignore my comment.

TBonnin:
I vote `scanM`, `queryM` and `ScanPaginatedM` / `queryPaginatedM`.
Up to you though

regiskuckaertz:
All done!

scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
```diff
@@ -43,6 +46,27 @@ private[scanamo] trait DynamoResultStream[Req, Res] {
...
+    req: Req,
+    pageSize: Int
+  )(implicit M: Alternative[M]): ScanamoOpsT[M, List[Either[DynamoReadError, T]]] = {
+    def streamMore(req: Req, l: Int): ScanamoOpsT[M, List[Either[DynamoReadError, T]]] =
+      exec(withLimit(l)(req)).toFreeT[M].flatMap { res =>
+        val results = items(res).map(ScanamoFree.read[T])
+        if (results.isEmpty)
+          FreeT.liftT(M.empty)
+        else {
+          val l1 = limit(req).map(_ - results.length).fold(pageSize)(Math.min(pageSize, _))
```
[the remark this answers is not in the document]
TBonnin: @regiskuckaertz have you seen this comment? ^^^

scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
TBonnin: @regiskuckaertz ^^^

scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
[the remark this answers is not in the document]
regiskuckaertz: ah no, sorry 🙈 lol. you're right!

scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
regiskuckaertz: No it shouldn't, let me try.

PR #434, opened 2019-06-12
Do not encode empty maps as null attribute values

### LINKED ISSUE #433: 1.0.0-M10 breaks Map.empty support
It looks like there is behavior change when it comes to storing `Map.empty` values in `1.0.0-M10`.

Previously (`1.0.0-M8` and prior), when you put a record with a `Map.empty` value, it would store it as

```
"name": {
  "M": {}
}
```

but with `1.0.0-M10`, it looks like the same value is stored as

```
"name": {
  "NULL": true
}
```

Unfortunately, with the same `1.0.0-M10`, if you try to get this record back out, you will get the following error:

```
'name': not of type: 'M' was 'DynNull'
```

regiskuckaertz:
Closes #433

### DIFF
--- a/formats/src/main/scala/org/scanamo/DynamoObject.scala
+++ b/formats/src/main/scala/org/scanamo/DynamoObject.scala
@@ -120,12 +120,7 @@ sealed abstract class DynamoObject extends Product with Serializable { self =>
   /**
     * Make an AWS SDK value out of this map
     */
-  // TODO: mark as private?
-  final def toAttributeValue: AttributeValue =
-    self match {
-      case Empty => DynamoValue.Null
-      case _     => new AttributeValue().withM(toJavaMap)
-    }
+  final def toAttributeValue: AttributeValue = new AttributeValue().withM(toJavaMap)
 
   /**
     * Builds a [[scala.collection.Map]] if this map is made entirely of values of type `V`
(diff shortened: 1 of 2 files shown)


formats/src/main/scala/org/scanamo/DynamoObject.scala
```diff
@@ -120,12 +120,7 @@ sealed abstract class DynamoObject extends Product with Serializable { self =>
   /**
     * Make an AWS SDK value out of this map
     */
-  // TODO: mark as private?
-  final def toAttributeValue: AttributeValue =
-    self match {
-      case Empty => DynamoValue.Null
-      case _     => new AttributeValue().withM(toJavaMap)
-    }
+  final def toAttributeValue: AttributeValue = new AttributeValue().withM(toJavaMap)
```
TBonnin: Let's add a test no?

regiskuckaertz: Yes 👍

formats/src/test/scala/org/scanamo/DynamoFormatTest.scala
```diff
@@ -44,4 +44,10 @@ class DynamoFormatTest extends FunSpec with Matchers with ScalaCheckDrivenProper
   testReadWrite[Set[String]](Gen.containerOf[Set, String](nonEmptyStringGen))
   testReadWrite[Option[String]](Gen.option(nonEmptyStringGen))
   testReadWrite[Option[Int]]()
+  testReadWrite[Map[String, Long]](Gen.mapOf[String, Long] {
+    for {
+      key <- nonEmptyStringGen
+      value <- Arbitrary.arbitrary[Long]
+    } yield key -> value
+  })
```
TBonnin: I think you can use `Gen.zip` rather than a for comprehension.
```
Gen.mapOf[String, Long](Gen.zip(nonEmptyStringGen, Arbitrary.arbitrary[Long]))
```

regiskuckaertz: I looked for that but couldn't find it 🤔

Issue #398, opened 2019-05-02
Review key, condition and update expressions


regiskuckaertz:
One of the key features of scanamo is the nice DSL for writing expressions. This is achieved via a combination of features of the Scala language that I don't think are very functional:

- implicit conversions and classes
- the fact that subclasses in a sealed trait are bona fide "types"
- operator overloading
- bounds to data types

The reason is that, for instance, [the operator `and`](https://github.com/scanamo/scanamo/search?q=%22def+and%22&unscoped_q=%22def+and%22) can mean different things in different settings. It can be a combination of filters or keys. To pull that off, internal "typeclasses" are defined to make an operation polymorphic, then data types are defined to contain values for which there exists an instance of those typeclasses. That is why we see things like `UniqueKey[_]` all over the place.

Let's try to rethink that approach, starting from the beginning, and address each problem separately. This may mean that syntactic sugar will be different in the end, but we can certainly keep a clean and simple API.

Areas to work on:

- replace `UniqueKey[T]` and `UniqueKeys[T]` with `Key` and `List[Key]`
```scala
sealed abstract class Key extends Product with Serializable {
  type V
  def toDynamoValues(implicit DynamoFormat[V]): Map[String, DynamoValue]
}

object Key {
  final case class Partition[A](n: String, p: A) extends Key {
    type V = A
  }
  final case class PartitionAndSort[A, B](p: (String, A), s: (String, B)) extends Key {
    type V = (A, B)
  }
}
```

- replace `ConditionExpression[T]` with `Condition`
```scala
sealed abstract class Condition extends Product with Serializable {
  type V
  def toRequestCondition(implicit DynamoFormat[V]): RequestCondition
}

object Condition {
  final case class Comparison(op1: Operand, cmp: Comparator, op2: Operand) extends Condition
  final case class Between(op: Operand, lo: Operand, hi: Operand) extends Condition
  ...
  final case class Contains(at: AttributeName, op: Operand) extends Condition
}

sealed abstract class Operand extends Product with Serializable
final case class Literal[A](l: A) extends Operand
final case class Attribute(a: AttributeName) extends Operand
final case class Size(a: AttributeName) extends Operand
```

PR #443, opened 2019-06-20
Introduce `Key`

regiskuckaertz:
Partially addresses #398 

The new `Key` replaces `UniqueKey` and has two type parameters: `Key[KT, T]`

- `KT` identifies whether the table has a `Simple` or `Composite` key
- `T` is the type of key values. In the `Composite` case, it is actually `(T', T")`

`KT` is just a phantom type to make sure a key is "correct by construction". This changes ripples through the whole API, notably `Table` is now indexed by these two types to make sure all APIs relying on keys receive correct values.

This is still a WIP as I want to tackle `Query` in here too, because now we can enforce the `query*` APIs can only be called when the table has a composite key at compile time, which I think is great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ scala> val farmersTableResult = LocalDynamoDB.createTable(client)("farmer")("nam
 
 scala> case class Farm(animals: List[String])
 scala> case class Farmer(name: String, age: Long, farm: Farm)
-scala> val table = Table[Simple, String, Farmer]("farmer")
+scala> val table = Table[String, Farmer]("farmer")
 
 scala> val ops = for {
      |   _ <- table.putAll(Set(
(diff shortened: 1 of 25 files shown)


TBonnin:
❤️ 0c4d3f0b98140b946c7a570ee074b6d3d1edb470
Why calling it `IsSimpleKey` instead of `SimpleKey`?

regiskuckaertz:
Naming, right 😆 `SimpleKey` is shorter 👍 I'm not sure about whether or not we should give access to it or keep it private. There's only a few simple types, but people might create types that map to them, e.g. `UUID -> String`. We can't currently represent that because `DynamoFormat.xmap` loses that information, but it would be to address that I think.

PR #451, opened 2019-06-28
RetryPolicy fixes

### LINKED ISSUE #448: Combining RetryStrategies broken in M10?
Hey Folks, 
thanks for your great work on this library so far 👍 
After updating to `1.0.0-M10`, I was trying out the new `RetryStrategy` from #402 and stumbled on this weird behaviour when combining strategies, being unable to combine delay with a maximum amount of retries.

```scala
import akka.actor.ActorSystem
import akka.stream.ActorMaterializer
import akka.stream.scaladsl.{Sink, Source}
import org.scalatest.{Assertion, BeforeAndAfterAll, FreeSpec}
import org.scanamo.ops.retrypolicy.{RetryPolicy, WithRetry}

import scala.concurrent.Await
import scala.concurrent.duration._
import scala.util.control.NonFatal

class ScanamoIssue extends FreeSpec with BeforeAndAfterAll with WithRetry {

  implicit val actorSystem: ActorSystem        = ActorSystem()
  implicit val materializer: ActorMaterializer = ActorMaterializer()

  override protected def afterAll(): Unit = {
    materializer.shutdown()
    Await.ready(actorSystem.terminate(), 10.seconds)
  }

  private val tries  = 5
  private val delay  = 50.millis
  private val factor = 2.0

  private def testcase(policy: RetryPolicy): Assertion = {
    val errors = Await.result(
      retry(Source.failed(new LimitExceededException("Limit Exceeded")), policy)
        .recover {
          case NonFatal(t) => t
        }
        .runWith(Sink.seq),
      5.seconds
    )
    assert(errors.size == 1)
  }

  private def testcaseWithDelay(policy: RetryPolicy) = {
    val minDelay = tries * delay.toMillis
    val start = System.currentTimeMillis()
    testcase(policy)
    val stop = System.currentTimeMillis()
    assert(stop - start > minDelay)
  }

  "Retry Policies should terminate" - {
    "Plain maximum" in testcase(RetryPolicy.Max(tries))
    "Constant delay AND maximum" in testcaseWithDelay(RetryPolicy.Max(tries) && RetryPolicy.Constant(delay))
    "Linear delay AND maximum" in testcaseWithDelay(RetryPolicy.Max(tries) && RetryPolicy.Linear(delay, factor))
    "Exponential delay AND maximum" in testcase(RetryPolicy.Max(tries) && RetryPolicy.Exponential(delay, factor))
    "Constant delay OR maximum" in testcaseWithDelay(RetryPolicy.Max(tries) || RetryPolicy.Constant(delay))
    "Linear delay OR maximum" in testcaseWithDelay(RetryPolicy.Max(tries) || RetryPolicy.Linear(delay, factor))
    "Exponential delay OR maximum" in testcase(RetryPolicy.Max(tries) || RetryPolicy.Exponential(delay, factor))
  }
}
```
I'd expect all of these tests to pass, but only the first one does. 
As far as I see it, there's currently no way of combining limited retries with any sort of delay, even though the API suggests this. 

Am I missing something here?

Cheers,
Joe

regiskuckaertz:
Closes #448 

Also steals tests from the ticket.

Fixes:
- Max bottoms out at 0 and never goes below
- switch `Duration` for `FiniteDuration` (`Never ~ Duration.Inf`)
- Fixes exponential delay calculation
- adds constructors to guarantee correctness
- proper introduction of a delay in the stream algebra

### DIFF
--- a/alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
+++ b/alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
@@ -5,30 +5,29 @@ import org.scanamo.ops.retrypolicy.RetryPolicy._
 import scala.concurrent.duration.FiniteDuration
 
 sealed abstract class RetryPolicy extends Product with Serializable { self =>
-  final def done: Boolean = self match {
-    case Max(1)                      => true
-    case And(thisPolicy, thatPolicy) => thisPolicy.done && thatPolicy.done
-    case Or(thisPolicy, thatPolicy)  => thisPolicy.done || thatPolicy.done
-    case Never                       => true
-    case _                           => false
+  final def continue: Boolean = self match {
+    case Max(0) | Never              => false
+    case And(thisPolicy, thatPolicy) => thisPolicy.continue && thatPolicy.continue
+    case Or(thisPolicy, thatPolicy)  => thisPolicy.continue || thatPolicy.continue
+    case _                           => true
   }
 
   final def delay: Long = self match {
-    case Constant(retryDelay)       => retryDelay.toMillis
-    case Linear(retryDelay, factor) => retryDelay.toMillis * factor.toLong
-    case Exponential(retryDelay, factor) =>
-      val delayMillis = retryDelay.toMillis
-      delayMillis * Math.pow(delayMillis.toDouble, factor).toLong
-    case And(thisPolicy, thatPolicy) => Math.max(thisPolicy.delay, thatPolicy.delay)
-    case Or(thisPolicy, thatPolicy)  => Math.min(thisPolicy.delay, thatPolicy.delay)
-    case _                           => 0
+    case Constant(retryDelay)            => retryDelay.toMillis
+    case Linear(retryDelay, factor)      => retryDelay.toMillis * factor.toLong
+    case Exponential(retryDelay, factor) => retryDelay.toMillis * factor.toLong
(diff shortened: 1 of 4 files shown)


alpakka/src/main/scala/org/scanamo/ops/retrypolicy/RetryPolicy.scala
```diff
@@ -2,53 +2,82 @@ package org.scanamo.ops.retrypolicy
...
-    case Or(thisPolicy, thatPolicy)  => Math.min(thisPolicy.delay, thatPolicy.delay)
-    case _                           => 0
+  final def delay: Duration = self match {
+    case Constant(retryDelay)               => retryDelay
+    case Linear(retryDelay, n)              => retryDelay * n.toDouble
+    case Exponential(retryDelay, factor, n) => retryDelay * Math.pow(factor, n.toDouble)
+    case And(thisPolicy, thatPolicy) =>
+      val d1 = thisPolicy.delay
+      val d2 = thatPolicy.delay
+      if (d1 >= d2) d1 else d2
```
TBonnin: `Duration` has `min/max` function: 

```
case And(thisPolicy, thatPolicy) => thisPolicy.delay.max(thatPolicy.delay)
```

regiskuckaertz: Oh great!

PR #460, opened 2019-07-11
Fix NULLs when writing empty lists

desbo:
As discussed with @regiskuckaertz.

A quick observation as a first-time contributor: it might be good to [configure the local Dynamo instance](https://github.com/localytics/sbt-dynamodb#configuration) to automatically launch and terminate when running tests.

### DIFF
--- a/formats/src/main/scala/org/scanamo/DynamoArray.scala
+++ b/formats/src/main/scala/org/scanamo/DynamoArray.scala
@@ -83,7 +83,7 @@ sealed abstract class DynamoArray extends Product with Serializable { self =>
     */
   // TODO mark as private?
   final def toAttributeValue: AttributeValue = self match {
-    case Empty       => new AttributeValue().withL()
+    case Empty       => emptyListAttributeValue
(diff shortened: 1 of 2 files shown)


formats/src/main/scala/org/scanamo/DynamoArray.scala
```diff
@@ -83,7 +83,7 @@ sealed abstract class DynamoArray extends Product with Serializable { self =>
     */
   // TODO mark as private?
   final def toAttributeValue: AttributeValue = self match {
-    case Empty       => DynamoValue.Null
+    case Empty       => new AttributeValue().withL()
```
regiskuckaertz: Can you cache that value as a private val on the companion?

regiskuckaertz:
> A quick observation as a first-time contributor: it might be good to configure the local Dynamo instance to automatically launch and terminate when running tests.

💯 I'll create a ticket, I think the SBT build can be improved in many ways, esp. the testing experience.

PR #468, opened 2019-07-22
Remove scalaz Interpreter

markus1189:
https://github.com/scalaz/ioeffect is no longer active and in favor of progressing towards Scala 2.13 https://github.com/scanamo/scanamo/issues/335 support is dropped


build.sbt
```diff
@@ -183,24 +181,6 @@ lazy val catsEffect = (project in file("cats"))
...
   .dependsOn(formats, scanamo, testkit % "test->test")
 
-lazy val scalaz = (project in file("scalaz"))
-  .settings(
-    name := "scanamo-scalaz",
-    commonSettings,
-    publishingSettings,
-    libraryDependencies ++= List(
-      awsDynamoDB,
-      "com.codecommit" %% "shims"           % shimsVersion,
```
regiskuckaertz: I think `shimsVersion` was only used here.

markus1189: Correct and removed :)

PR #469, opened 2019-07-26
Add generic set to Dynamo Formats

ajwl:
This allows for a Generic Scala `Set` to be interpreted by the program. 

Previously Sets of numbers and strings were accepted, but not Sets of custom objects. Provided the items are in a DynamoFormat, this can now be done.


ajwl:
@regiskuckaertz !

PR #481, opened 2019-08-10
Minor updates

regiskuckaertz:
- updates ZIO to latest version, remove traces of scalaz
- moves the java time project into core


regiskuckaertz:
It was a mistake. Instances for data types that are part of the standard library should be readily available, otherwise it becomes annoying hunting down which imports to use.

PR #484, opened 2019-08-19
Remove DynamoFormat[A].default

regiskuckaertz:
To handle some of the [quirks of DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html):

> Attribute values cannot be null. String and Binary type attributes must have lengths greater than zero. Set type attributes cannot be empty. Requests with empty values will be rejected with a ValidationException exception.

The `default` value was introduced in `DynamoFormat`; but it doesn't belong there. Even though nulls are forbidden, they get first-class representation in the `AttributeValue` type, and it is also trivial to represent empty sets or all the other cases. Instead we can push that handling down into the (un-)marshalling logic.

The only thing I'm not sure of is whether to change
```scala
trait DynamoObject {
  def apply(key: String): Option[DynamoValue]
}
```
into
```scala
trait DynamoObject {
  def apply(key: String): DynamoValue
}
```
and treat missing properties as `DynamoValue.nil`.


formats/src/main/scala/org/scanamo/DerivedDynamoFormat.scala
```diff
@@ -28,30 +28,28 @@ trait DerivedDynamoFormat {
...
-            .map(headFormat.value.read)
-            .orElse(headFormat.value.default.map(Either.right))
-            .getOrElse(Either.left(MissingProperty))
-            .leftMap(((fieldName, _)))
-
-        val head = valueOrError.toValidatedNel.map(field[K](_))
-        val tail = tailFormat.value.read(av)
+      final def read(av: DynamoObject): ValidatedPropertiesError[FieldType[K, V] :: T] = {
+        val head = av(fieldName)
+          .fold(headFormat.read(DynamoValue.nil).leftMap[DynamoReadError](_ => MissingProperty))(headFormat.read(_))
```
regiskuckaertz: For reviewers: this is where I am passing `nil` explicitly if the key is missing from the object. If you take a look at the formats for which there should be a default value, they always test if the value is null and return the default value in that case. Two things:

1. `apply` returning an `Option[DynamoValue]` allows me to isolate the case of a `MissingProperty`
2. OTOH knowing that dynamo will never return a null value, if `apply` returns `DynamoValue` instead, I can infer the `MissingProperty` error if `av(fieldName).isNull`

regiskuckaertz:
Yes, there are property-based tests around that specific feature.

I tend to agree, we just need to make sure this won't be a problem when manually writing instances using `value.as[Option[SomeClass]]` or similar.

YAY!

PR #485, opened 2019-08-19
Remove simulacrum

regiskuckaertz:
One set of macros we can do without

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/query/UniqueKeyCondition.scala
+++ b/scanamo/src/main/scala/org/scanamo/query/UniqueKeyCondition.scala
@@ -3,9 +3,8 @@ package org.scanamo.query
 import cats.instances.option._
 import cats.syntax.apply._
 import org.scanamo.{ DynamoFormat, DynamoObject }
-import simulacrum.typeclass
 
-@typeclass trait UniqueKeyCondition[T] {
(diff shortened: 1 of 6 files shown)


scanamo/src/main/scala/org/scanamo/query/UniqueKeyCondition.scala
```diff
@@ -3,50 +3,52 @@ package org.scanamo.query
...
-@typeclass trait UniqueKeyCondition[T] {
+trait UniqueKeyCondition[T] {
   type K
   def toDynamoObject(t: T): DynamoObject
   def fromDynamoObject(key: K, dvs: DynamoObject): Option[T]
   def key(t: T): K
 }
 
 object UniqueKeyCondition {
+  type Aux[T, K0] = UniqueKeyCondition[T] { type K = K0 }
```
regiskuckaertz: not sure why I had to reify the type here, there is nothing in the simulacrum macro for that. perhaps a side-effect of macro expansion happening at a different time in the compilation process.

regiskuckaertz: (`UniqueKeyCondition` will eventually go away, so it's not really an issue)

PR #486, opened 2019-08-21
Add `putAndReturn`, use `Unit` as return type

### LINKED ISSUE #464: Put succeeds but always Returns None
When using the latest M10 performing a basic PUT operation, it always returns None for the Option[Either[]], The expected result is a Some(Right()) since manual inspection has verified that the record is written to the LocalDynamoDB instance.

```
main: /Users/stevenfines/workspace/user-service/data/test/shared-local-instance.db
sqlite> select ObjectJSON from "test_tvp_users_api";
{"lastName":{"S":"User"},"applicationRole":{"S":"unittest"},"clientId":{"N":"1"},"createdTimestamp":{"N":"1563318931312"},"groups":{"NULL":true},"userSlug":{"S":"test_user"},"incentiveHistory":{"NULL":true},"autoApprove":{"BOOL":false},"firstName":{"S":"Test"},"permissions":{"NULL":true},"modifiedTimestamp":{"N":"1563318931312"},"id":{"N":"1"},"activeNotifications":{"NULL":true},"customAttributes":{"NULL":true}}
sqlite> 
```

Here's the code example:
```
def createUserRecord(user:User): Boolean = {
    val inserted = concurrent.Await.result(ScanamoAsync(provider.getClient).exec(table.put(user)), 5.seconds)

    if(logger.isErrorEnabled()){
     logger.error(s"Operation Result: $inserted,\nOperation Successful: ${inserted.isDefined && inserted.get.isRight}")
    }

    inserted.fold(false){
      case Left(err: DynamoReadError) => {
        logger.error(DynamoReadError.describe(err))
        false
      }
      case Right(user: User) => {
        true
      }
      case Left(x) => {
        logger.error(x.toString)
        false
      }
    }
  }
```

regiskuckaertz:
Closes #464


regiskuckaertz:
Hold on before reviewing, I want to change this to something like:
```scala
case class Returned[A](
  itemCollectionMetrics: Option[ItemCollectionMetrics],
  consumedCapacity: Option[ConsumedCapacity],
  attributes: Option[DynamoValue]
)

val table = Table[Person]
for {
  ret <- table.returning(Return.OldValue).delete(id)
} yield ret.attributes.map(_.as[Person])
```

PR #491, opened 2019-08-27
Fix package name and version for scanamo.org index page

### LINKED ISSUE #489: Package imports names do not match between github and scanamo.org
There are conflicting package names given for Scanamo.

https://github.com/scanamo/scanamo/blob/master/README.md specifies:
`"org.scanamo" %% "scanamo" % "1.0.0-M10"`

However https://www.scanamo.org/ specifies:
`"com.gu" %% "scanamo" % "1.0.0-M8"`

I'm guessing that the github version is the right one to choose, but it's confusing to have both.

TBonnin:
Fix #489


TBonnin:
This should be automated to be honest. 😆

PR #511, opened 2019-09-19
Add support for 2.13

### LINKED ISSUE #335: Build for scala 2.13

regiskuckaertz:
- main scala version is now 2.12
- cross-builds for 2.13 and 2.11 (I'm thinking of dropping the latter ... opinion?)
- dropped the toolbox dependency as it wasn't super useful

Closes #335


build.sbt
```diff
@@ -1,5 +1,5 @@
-scalaVersion in ThisBuild := "2.12.8"
-crossScalaVersions in ThisBuild := Seq("2.11.12", scalaVersion.value)
+scalaVersion in ThisBuild := "2.12.10"
```
rtyley: The PR description says:

> * main scala version is now 2.13

...so I would expect...

```suggestion
scalaVersion in ThisBuild := "2.13.1"
```

...or happy to stick with `2.12.10`for now if that is the case?

Have no problem at all with dropping Scala 2.11 by the way!

regiskuckaertz: Haha, you are right. I shifted gears because one plugin is misbehaving and have just updated the description accordingly. Also on second thought, until 2.13 is the version used by the majority of users, it may be preferable for the library to keep using 2.12 for dev. What do you think?

rtyley: As they're both getting built by CI, I think I'd favour whichever gave the best developer experience - if 2.13 is a faster _compile_ for instance, that would be a good point in its favour, but if a plugin is misbehaving on 2.13 and that will negatively affect contributors, probably best avoided for now.

regiskuckaertz: Ok, agreed. It is definitely the fastest compiler. I'll revisit in the future, another option (since this plugin is for doc generation) is that we only build docs for 2.12. That's what the author suggests.

regiskuckaertz:
@johnduffell Thanks 🙏  I had removed the coverage plugin for the same reason but just restored it.

PR #538, opened 2019-10-25
Generic derivation with Magnolia

### LINKED ISSUE #138: Enum value types mapped incorrectly when used directly in queries

### LINKED ISSUE #316: Compilation error when too much field in case class
Hello, 

I use scanamo version 1.0.0M1 (I tried also on on M8 and M9...)
I've got a case class with 35 fields, when I add a field it doesn't compile anymore, i've got on error on missing implicit DynamoFormat. I get this with "sbt compile"

But when I test my Dao in one shot class in intellij, it works.

Is there a hard limit concerning the "size" of the case class ?

my case class only contains fields of Option[Double] and Option[String], no exotic objects....

thanks

Philippe

### LINKED ISSUE #495: With Dyanmic format enabled, compile takes 30+ minutes and then crashes with a GC Overhead exception
With these Dynamo Formats:
```scala
object DynamoFormats {
  import org.scanamo.DynamoFormat._

  implicit val timestampDateTimeFormat = coercedXmap[Timestamp, Long, IllegalArgumentException](x => new java.sql.Timestamp(x))(_.getTime)
  implicit val dateFormat = coercedXmap[Date, Long, IllegalArgumentException](x => new Date(x))(_.getTime)
  implicit def setFormat[T](implicit f: DynamoFormat[T]): DynamoFormat[Set[T]] = xmap[Set[T], List[T]](l => Either.right(l.toSet))(_.toList)
  implicit def uriFormat = coercedXmap[URI,String, IllegalArgumentException](x => URI.create(x))(_.toASCIIString())
  implicit def incentiveTypeFormat= coercedXmap[IncentiveType.Value, String, IllegalArgumentException](x => IncentiveType.withNameOpts(x).get)(_.toString())

  implicit val incentiveHistoryFormat:DynamoFormat[IncentiveHistory] = deriveDynamoFormat[IncentiveHistory]
  implicit val groupFormat = deriveDynamoFormat[Group]
  implicit val permissionFormat = deriveDynamoFormat[Permission]
  implicit val notificationFormat = deriveDynamoFormat[Notification]
  implicit val clientCustomFormat = deriveDynamoFormat[ClientCustomAttribute]
  implicit val locationFormat = deriveDynamoFormat[LocationInfo]
  implicit val userFormat = deriveDynamoFormat[User]
  implicit val accountUserFormat = deriveDynamoFormat[Account]
}
```
[User.scala.txt](https://github.com/scanamo/scanamo/files/3552741/User.scala.txt)

The compiler will take 30+ minutes and then finally fall over with a stack overflow exception. I've dug through them (the case classes are defined in the attachment) and I don't see any recursive elements which would cause this issue- the vast majority of the time is spent in the phase typer phase on the object above

Compiler is: scala 2.12.8, running on Oracle's JDK 1.8.0_212 on mac os macOS 10.14.6 (18G87)

regiskuckaertz:
Closes #215 
Closes #316 
Closes #495 
Closes #138 

This is a big PR, but I couldn't find an easier way around it. List of changes:

- Replaces shapeless with magnolia for faster generic derivation. It also improves on it, e.g. we are  now able to handle case class fields with default values. There are more which can be done, we can explore in the future
- Removes the format project: this made it easier for me to work. It is not ideal, we don't want to release sync and async interpreters in alpakka, etc. This will be cleaned up in a follow-up. But fundamentally, `DynamoFormat` is central to scanamo so there was no reason to keep it separate. However, the generic stuff can be moved away if needed.
- Move the generic stuff into the `generic` package, it feels slightly better.
- Moves `DynamoReadError` into the core package, again it is a core data type, no need to put into a subpackage.
- Fixes an issue where decoding a `List[X]` would not work for strings, numbers and bytebuffers. The fix is not perfect, but can be improved in a follow-up.
- Changes the encoding of sealed traits to an object with two fields: `tag` identifies the branch and `value` is the elaboration of the branch into a `DynamoValue`. This makes decoding easier at the cost of some duplication for case objects.

The interesting stuff really is the magnolia derivation, which can be understood by reading [the tutorial](https://propensive.com/opensource/magnolia/tutorial).


cats/src/test/scala/org/scanamo/fixtures/fixtures.scala
```diff
@@ -0,0 +1,18 @@
...
+case class Farm(asyncAnimals: List[String])
+case class Farmer(name: String, age: Long, farm: Farm)
+case class Forecast(location: String, weather: String, equipment: Option[String])
+case class Gremlin(number: Int, wet: Boolean)
+case class Item(name: String)
+case class Lemming(name: String, stuff: String)
+case class Rabbit(name: String)
+case class Station(mode: String, name: String, zone: Int)
+case class Transport(mode: String, line: String, colour: String)
+case class Worker(firstName: String, surname: String, age: Option[Int])
```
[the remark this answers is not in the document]
regiskuckaertz: You're right, I should do that 🤦‍♂ Another thing I wondered is whether generic derivation should be in its own package. I guess it depends on three things:
- manual derivation should have minimal boilerplate
- generic derivation should add a significant cost to the build (either in terms of size or speed)
- all tests using auto are converted to manual

Right now I think 1 is definitely false, 2 may not be either with this change (I will test with the CAPI data types), and 3 may be a significant effort.

regiskuckaertz:
One further refactoring I'd like to do is abstract the test script over the interpreter. Since they're all the same, I don't think this is too hard to achieve, but something for a follow-up.

regiskuckaertz:
I'm quite happy with this empirical benchmark.

## Before (another PR)

Screen Shot 2019-11-08 at 06 59 44

## After (this PR)

Screen Shot 2019-11-08 at 06 58 15

scanamo/src/main/scala/org/scanamo/generic/Derivation.scala
```diff
@@ -0,0 +1,94 @@
...
+        def writeObject(t: T): DynamoObject =
+          DynamoObject(cc.parameters.foldLeft(List.empty[(String, DynamoValue)]) {
+            case (xs, p) =>
+              val v = unbuild(p.typeclass).write(p.dereference(t))
+              if (v.isNull) xs else (p.label -> v) :: xs
+          }: _*)
+      })
+  }
+
+  // Derivation for ADTs, they are encoded as an object of two properties:
```
regiskuckaertz: Open to debate this.

scanamo/src/main/scala/org/scanamo/DynamoObject.scala
```diff
@@ -21,11 +24,14 @@ sealed abstract class DynamoObject extends Product with Serializable { self =>
     */
   final def apply(key: String): Option[DynamoValue] = self match {
     case Empty          => None
-    case Strict(xs)     => Option(xs.get(key)).map(DynamoValue.fromAttributeValue)
+    case Strict(xs)     => val x = xs.get(key); if (x eq null) None else Some(DynamoValue.fromAttributeValue(x))
```
TBonnin: Isn't `Option().map` the same?

scanamo/src/main/scala/org/scanamo/generic/Derivation.scala
TBonnin: Do we need this? Scanamo doesn't have this feature at the moment right?

alpakka/src/test/scala/org/scanamo/ScanamoAlpakkaSpec.scala
```diff
@@ -516,9 +474,8 @@ class ScanamoAlpakkaSpec extends FunSpecLike with BeforeAndAfterAll with Matcher
 
   it("should get multiple items asynchronously (automatically handling batching)") {
     LocalDynamoDB.usingRandomTable(client)("id" -> N) { t =>
-      case class Farm(id: Int, name: String)
-      val farms = (1 to 101).map(i => Farm(i, s"Farm #$i")).toSet
-      val farmsTable = Table[Farm](t)
+      val farms = (1 to 101).map(i => Factory(i, s"Farm #$i")).toSet
+      val farmsTable = Table[Factory](t)
```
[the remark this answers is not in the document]
regiskuckaertz: Yes, there is another definition of `Farm` that is different from the one used in that test, so I created another class as a quick fix 😋

scanamo/src/main/scala/org/scanamo/DynamoObject.scala
regiskuckaertz: Semantically, yes. Operationally, there is less allocation (the initial `Option(xs.get(key))` is created on the heap and thrown away immediately).

scanamo/src/main/scala/org/scanamo/generic/Derivation.scala
regiskuckaertz: It's [here](https://github.com/scanamo/scanamo/blob/master/formats/src/main/scala/org/scanamo/DerivedDynamoFormat.scala#L67-L92). At the moment it expects an object like this:
```
{
  "<className>": <value>
}
```
but I changed it to:
```
{
  "tag": "<className>"
  "value": <value>
}
```
Now that I look at it, I am not sure it is better 🤔

regiskuckaertz:
It doesn't break anything, no. Semi- and auto-derivation work as usual, except they will compile much faster. I'll add some comments to the code, it is hard to read if you don't actually dig into the code of Magnolia.

regiskuckaertz:
Thanks for the review 🙏

scanamo/src/main/scala/org/scanamo/generic/Derivation.scala
regiskuckaertz: ... and it would break existing store values. So I'll change that.

regiskuckaertz:
Fixed. I've also added a new feature: fields with default values are now properly handled. That wasn't possible with Shapeless 🎉

TBonnin:
Awesome job @regiskuckaertz

PR #562, opened 2019-12-10
Transactional writes

semenodm:
Only Put supported now, without conditions.
Multiple table transactional writes


scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
```diff
@@ -37,6 +37,24 @@ object ScanamoFree {
         ScanamoOps.batchWrite(new BatchWriteItemRequest().withRequestItems(map))
       }
 
+  def transactWriteToTable[T](
```
regiskuckaertz: Same here, `transactPutAllTable`

scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
```diff
@@ -37,6 +37,24 @@ object ScanamoFree {
...
+    tableName: String
+  )(items: List[T])(implicit f: DynamoFormat[T]): ScanamoOps[TransactWriteItemsResult] =
+    transactWrite(List.fill(items.size)(tableName).zip(items))
+
+  def transactWrite[T](
```
regiskuckaertz: Same here.

scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
```diff
@@ -37,6 +37,24 @@ object ScanamoFree {
         ScanamoOps.batchWrite(new BatchWriteItemRequest().withRequestItems(map))
       }
 
+  def transactWriteToTable[T](
+    tableName: String
+  )(items: List[T])(implicit f: DynamoFormat[T]): ScanamoOps[TransactWriteItemsResult] =
+    transactWrite(List.fill(items.size)(tableName).zip(items))
```
regiskuckaertz:
```suggestion
    transactWrite(items.map(tableName -> _))
```

scanamo/src/main/scala/org/scanamo/Table.scala
```diff
@@ -796,6 +802,9 @@ case class Table[V: DynamoFormat](name: String) {
 
   def descending =
     TableWithOptions(name, ScanamoQueryOptions.default).descending
+
+  def transactWriteToTable(vs: List[V]): ScanamoOps[TransactWriteItemsResult] =
```
regiskuckaertz: Let's call it `transactPutAll` for now, as I'd like to reserve the `Write` for when we support the full gamut of operations.

.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 *dynamodb-local
+.idea/
```
regiskuckaertz: This should be in your personal .gitignore instead: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

PR #569, opened 2019-12-27
Dotty build

regiskuckaertz:
Will bring your PR in line with the latest changes and also cross-build in CI.


regiskuckaertz:
Gonna merge this into your branch @shtukas . Two remaining steps: 
- disable (semi-)automatic derivation for now in Dotty, as the story for macros is not clear yet. This can be achieved by moving the generic stuff into version specific folders, e.g. `src/main/scala-2.x/org/scanamo/...` and configuring the build appropriately
- solve that `Monad` requirement on the sync interpreter

PR #577, opened 2020-01-14
Fix delay in retry policy

Divyanshu564:


alpakka/src/main/scala/org/scanamo/ops/retrypolicy/WithRetry.scala
```diff
@@ -33,7 +33,7 @@ trait WithRetry {
           if (retryPolicy.continue) {
             Source
               .single(())
-              .delay(FiniteDuration(retryPolicy.delay, TimeUnit.MILLISECONDS))
+              .delay(FiniteDuration(retryPolicy.delay.toMillis, TimeUnit.MILLISECONDS))
```
regiskuckaertz: Notice that here, since `Duration.Inf` is possible, we may want to add:
```scala
trait RetryPolicy {
  def delayOrElse(or: => A)(f: FiniteDuration => A): A = self.delay match {
    case Duration.Inf => or
    case otherwise: FiniteDuration => f(otherwise)
  }
}
```

Divyanshu564: @regiskuckaertz could you please review it again

Divyanshu564:
@regiskuckaertz please merge this into master

PR #578, opened 2020-01-15
Only depend on scalatestplus-scalacheck in Test

benwaffle:
This was conflicting with a different version of scalatest, when depending on scanamo.


benwaffle:
awesome! when do you think this will be released?

regiskuckaertz:
I can release a M12-1 or something like that if blocking for you?

benwaffle:
Yes, Could you do that please?

kailuowang:
@regiskuckaertz a M12-1 release will be much appreciated here too, we are blocked by binary incompatibility caused by this (our test, due to scalatestplus-play is stuck with scalatest 3.0.8) 
Thanks!!!

kailuowang:
Thanks

PR #584, opened 2020-01-28
Allow customization of retryable exceptions

markus1189:
Previously it was hardcoded what `Throwable` should be retried (see
`WithRetry` trait).  However we are currently experiencing serious
issues with 500 from DynamoDB that we (in our case) would like to
retry.

With this pr, we pass a predicate `Throwable => Boolean` from
`ScanamoAlpakka` down to the `AlpakkaInterpreter` to implement the new
abstract `retryable` method from `WithRetry`.

By default, we use the same retry behavior as before, see
`ScanamoAlpakka.defaultRetryableCheck`

What do you think?


regiskuckaertz:
Thanks a lot for your contribution 🎉 🎉 🎉 🎉

PR #586, opened 2020-01-30
Change how SemiAuto- and AutoDerivation works

### LINKED ISSUE #583: Weird behavior in DynamoFormat derivation involving Option and Map[A,B]
Hey :wave:,

while migrating one of our services from scanamo M8 to M12 we noticed a very strange issue.  The auto-derived instances differ: In M8 the encoded `AttributeValue` looks good, but in M12, somehow the `Option` is encoded in the *wrong* way.

Here is a minimal example I managed to produce:

```scala
      case class OptionWrapper(value: Option[MapWrapper])
      case class MapWrapper(m: Map[String, String])

      val value = OptionWrapper(Some(MapWrapper(Map())))

      DynamoFormat[OptionWrapper].write(value)
        .toAttributeValue // only needed in M12
        .toString should ===("")
```

Under `M8` the result is
```
"{M: {value={M: {m={M: {},}},}},}"
```

While under `M12` the result is (notice the `Some=`):
```
"{M: {value={M: {Some={M: {value={M: {m={M: {},}},}},}},}},}"
```

This seems to only happen if there is a `Map` inside the `MapWrapper`, if you replace the type of `m` with something else the Option is encoded correctly

Any ideas on this? This is breaking our data model of existing items in dynamo :sweat_smile:

markus1189:
The problem was, that the derivation would (sometimes) prioritise the
derived instance instead of the one with higher priority in the
companion.

See the `AutoDerivationTest` for an example.

The fix is calling the Magnolia.gen macro directly in both derivation
traits and removing the whole `Exported` and `Hidden` logic.

See also: #583


markus1189:
Ugh will look at the tests later...

markus1189:
Hmm I got it to work, but the whole implicit priorisation story is a bit fiddly.

I kind of hacked around it in 48f6e4ae33e64555596fc2d1d864649e39263e78, would be awesome if you could critically have a look and maybe you can find a better way than essentially duplicating all the `DynamoFormat` instances?

scanamo/src/main/scala/org/scanamo/generic/AutoDerivation.scala
```diff
@@ -17,16 +17,13 @@
...
 
 import magnolia.Magnolia
-import org.scanamo.DynamoFormat
+import org.scanamo.{ DynamoFormat, DynamoFormatDefaultInstances }
 
 import scala.language.experimental.macros
 
-trait AutoDerivation extends Derivation {
-  type Typeclass[A] = ExportedDynamoFormat[A]
+trait AutoDerivation extends Derivation with DynamoFormatDefaultInstances with LowPriority
```
regiskuckaertz: Ha, yeah this is what I was trying to avoid. I wonder why Magnolia is not picking up the instance from the `DynamoFormat` companion 🤔

regiskuckaertz:
Thank you so much for your work on this. I don't have the mental bandwidth right now to go through it but will definitely this weekend.

ducanhkhu-ovo:
hi, is there any plan to merge this PR and release a new version to fix this bug? thanks

regiskuckaertz:
hi 👋 sorry for the lack of reply so far, I was busy during my last week at the guardian. I just want to take a look at the magnolia macro because it feels like a bug there, not here, in my opinion. If it doesn't respect "priority" of implicit resolution then I'm really surprised nobody raised the issue before. The situation I would like to avoid is when both `DynamoFormat` and `AutoDerivation` are in scope then this leads to more conflicts during implicit resolution.

scanamo/src/main/scala/org/scanamo/generic/SemiautoDerivation.scala
```diff
@@ -18,18 +18,9 @@ package org.scanamo.generic
 
 import magnolia.Magnolia
 import org.scanamo.DynamoFormat
-import scala.language.experimental.macros
 
-final private case class Hidden[A](instance: A) extends AnyVal
+import scala.language.experimental.macros
 
 trait SemiautoDerivation extends Derivation {
-  type Typeclass[A] = HiddenDynamoFormat[A]
```
joroKr21: Wasn't sure if I should open another issue about that but this `HiddenDynamoFormat` wasn't actually hidden.
So then this would happen:

```scala
import org.scanamo.DynamoFormat
import org.scanamo.generic.semiauto._

object ScanamoTest extends App {
  class Foo(i: Int, s: String)
  object Foo {
    // This is a self-reference!
    implicit val dynamoFormat: DynamoFormat[Foo] = deriveDynamoFormat
  }

  println(Foo.dynamoFormat) // null
}
```

scanamo/src/main/scala/org/scanamo/generic/AutoDerivation.scala
joroKr21: It's normal Scala, there are two places it looks for implicits:
  1. The regular scope (locals members, inherited members, imports)
  2. The implicit scope (companion objects, package objects, etc.)

in that order, so if it finds something in the regular scope it never even looks in the implicit scope.

And by importing `org.scanamo.generic.auto._` we provide an implicit in regular scope that can derive an instance for any sealed trait (including `Option`).

regiskuckaertz: Oooooh yeah, so importing the contents of the `DybamoFormat` companion into the derivation trait may fix the problem with `embed`. I’ll try that!

regiskuckaertz:
@markus1189 I found a way to avoid the duplication, based on the changes you made. It involves creating a dummy implicit that is only visible when the contents of the auto package is in lexical scope, something like:

```scala
private[scanamo] final case class Dummy(unit: Unit) extends AnyVal

trait AutoDerivation {
  // ^ no need to extend Derivation here
  final implicit val dummy: Dummy = Dummy(())
}
```

Then the `DynamoFormat` companion must extend `Derivation` and call the macro:

```scala
object DynamoFormat extends generic.Derivation {
  ...

  implicit final def genericFormat[A](implicit U: generic.Dummy): DynamoFormat[A] =
    macro deriveGenericFormat[A]

  def deriveGenericFormat[A: c.WeakTypeTag](c: scala.reflect.macros.whitebox.Context)(U: c.Tree): c.Tree = {
    val _ = U
    Magnolia.gen[A](c)
  }
}
```

Notice, to unlock full-auto, the `Dummy` instance must be in scope. This isn't ideal since I wanted to move the generic stuff into a separate project, but so far this is the only solution I found that doesn't involve shadowing duplicate definitions 😞

markus1189:
Wow that's creative. The dummy implicit feels like a hack, but there are already other patterns with `DummyImplicits` and here it get's the job done :man_shrugging:

joroKr21:
That's an interesting idea 👍

regiskuckaertz:
If you can update your branch with this I’m happy to merge it ✌️

markus1189:
Alright, I called it `AutoDerivationUnlocker` instead of `Dummy` :)  But otherwise it's done and seems to work

markus1189:
Important: this does not yet touch the encoding of the different cases of an ADT/sealed trait.

regiskuckaertz:
Looks like you just need to sort out the `Foobar` stuff and run scalafmt, then it'll be good to go.

scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
```diff
@@ -215,6 +201,30 @@ object DynamoFormat extends LowPriorityFormats {
   def coercedXmap[A, B: DynamoFormat, T >: Null <: Throwable: ClassTag](read: B => A, write: A => B): DynamoFormat[A] =
     xmap(coerce[B, A, T](read), write)
 
+  implicit final def genericFormat[A](implicit U: AutoDerivationUnlocker): DynamoFormat[A] = macro deriveGenericFormat[A]
+
+  def deriveGenericFormat[A: c.WeakTypeTag](c: scala.reflect.macros.whitebox.Context)(U: c.Tree): c.Tree = {
+    val _ = U
+    Magnolia.gen[A](c)
+  }
+
```
regiskuckaertz: I think source order does matter especially with vals, it may be preferrable to move these two definitions down at the bottom.

scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
```diff
@@ -530,7 +543,7 @@ object DynamoFormat extends LowPriorityFormats {
     * }}}
     */
   implicit def mapFormat[V](implicit f: DynamoFormat[V]): DynamoFormat[Map[String, V]] =
-    xmap[Map[String, V], DynamoObject](_.toMap[V], m => DynamoObject(m.toSeq: _*))(javaMapFormat)
+    DynamoFormat.xmap[Map[String, V], DynamoObject](_.toMap[V], m => DynamoObject(m.toSeq: _*))(javaMapFormat)
```
regiskuckaertz: These can be reverted now.

scanamo/src/main/scala/org/scanamo/generic/AutoDerivation.scala
```diff
@@ -16,17 +16,12 @@
 
 package org.scanamo.generic
 
-import magnolia.Magnolia
-import org.scanamo.DynamoFormat
+/**
+  * This class enables the automatic derivation of a [[org.scanamo.DynamoFormat]] instance if in lexical scope.
+  * The actual derivation happens inside the companion
+  */
+private[scanamo] class AutoDerivationUnlocker
```
regiskuckaertz:
```suggestion
private[scanamo] sealed abstract class AutoDerivationUnlocker
```

regiskuckaertz:
Really glad to see this build/unbuild crap go, thanks for doing this!

scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
markus1189: Done in 832120b183652b470ee07c6909505d473de2eb1f

scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
markus1189: Done in bf3175964a839fa7565cd9d3c978c071809bcbc5

scanamo/src/main/scala/org/scanamo/generic/AutoDerivation.scala
markus1189: Done in ee1f8e53186959626425e38ab4bca4d179d60ff6

markus1189:
The only build that fails is the one for 2.11 and Magnolia itself does not support 2.11... (https://github.com/propensive/magnolia/issues/171)

Is there a reason to keep 2.11 support in Scanamo?

regiskuckaertz:
It may be too early to drop 2.11 as it is still heavily used at the guardian, but yeah that's why I have a version check in the build to used an older magnolia for 2.11. I don't see why it is failing right now, nothing much has changed 🤔

joroKr21:
Please don't use Magnolia for 2.11 😄 It's surely a really old version full of bugs.
The reason we can't support Magnolia on 2.11 are lazy vals and they are needed for recursive data types.

regiskuckaertz:
Done, let's rebase this one.

markus1189:
Rebasing is a mess, because master was already merged into it at least once.  

I would suggest a squash merge instead, because all of the intermediate commits are not needed in this case (they are too fine granular anyway) :)

regiskuckaertz:
Great work! And thanks a lot for your contribution 🕺 💪

PR #587, opened 2020-01-30
Annotate return types

markus1189:
Some methods did not have return type
annotations (e.g. DynamoFormat.xmap) which meant that something
inference / tooltips would show a ugly type instead.  Fixed by adding
the annotations.


regiskuckaertz:
Oh yes! I wish there was a compiler flag to error when types are missing. It looks like you just need to run `scalafmt` and then it's ready to go 👍

markus1189:
Oh, I'm so used to scalafmt being run on every compile, I didn't even check :sweat_smile:  Any reason why it's not enabled in scanamo?

regiskuckaertz:
Travis has been flaky recently, this is super annoying.

PR #588, opened 2020-01-30
Add DynamoFormat#{xmap,coercedXmap,iso}

markus1189:
Adds the `xmap`/`coercedXmap`/`iso` method to all `DynamoFormat`, being essentially a
shortcut to the static `DynamoFormat.{xmap,coercedXmap,iso}`

Besides convenience, the type-inference is also much better.

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
+++ b/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
@@ -89,6 +89,11 @@ trait DynamoFormat[T] {
   def read(av: DynamoValue): Either[DynamoReadError, T]
   def read(av: AttributeValue): Either[DynamoReadError, T] = read(DynamoValue.fromAttributeValue(av))
   def write(t: T): DynamoValue
+
+  def iso[U](r: T => U, w: U => T): DynamoFormat[U] = DynamoFormat.iso(r, w)(this)
+  def xmap[U](r: T => Either[DynamoReadError, U])(w: U => T): DynamoFormat[U] = DynamoFormat.xmap(r)(w)(this)
+  def coercedXmap[U](read: T => U)(write: U => T): DynamoFormat[U] =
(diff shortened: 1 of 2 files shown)


scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
```diff
@@ -89,6 +89,11 @@ trait DynamoFormat[T] {
   def read(av: DynamoValue): Either[DynamoReadError, T]
   def read(av: AttributeValue): Either[DynamoReadError, T] = read(DynamoValue.fromAttributeValue(av))
   def write(t: T): DynamoValue
+
+  def iso[U](r: T => U, w: U => T): DynamoFormat[U] = DynamoFormat.iso(r, w)(this)
+  def xmap[U](r: T => Either[DynamoReadError, U])(w: U => T): DynamoFormat[U] = DynamoFormat.xmap(r)(w)(this)
```
regiskuckaertz: Let's be consistent and use a single parameter block here too.

regiskuckaertz: ... and if you dont mind applying the same in the companion, i.e. a single parameter block for both `xmap` and `coercedXmap`, that would be 🙏

markus1189: Did it in f680771cf355f2c54934a8b0fbb7bb815f040527

regiskuckaertz:
Small conflict to resolve then it's good to go, fantastic!

markus1189:
Hopefully good now :wink:

regiskuckaertz:
do you get the same error locally? been trying to run the build without success 😭

regiskuckaertz:
it's unrelated to this PR anyway so will merge.  Thanks  again for  your  awesome work!

PR #593, opened 2020-02-06
Temp Fix #589 - "Throughput exceeds the current capacity of your table or …

### LINKED ISSUE #589: Throttling: "Throughput exceeds the current capacity of your table or index" with async client and On-Demand mode
Hi. First of all thanks for providing this great library with a very clean and easy-to-use API.

I just started using DynamoDB for some batch insert workload. Since I don't know my capacity requirements yet, I chose the On-Demand mode. However, I can't seem to prevent write loss.

I'm using the async version of the API in conjunction with Alpakka DynamoDB `1.1.2`:

```
val asyncDynamoDbClient = DynamoClient(DynamoSettings(actorSystem))
```
1) when using the latest release `1.0.0-M12-1`:

```
val sc = ScanamoAlpakka(asyncDynamoDbClient, RetryPolicy.exponential(500.milliseconds, 2.0))
val result: Future[Unit] = sc.exec(rawSkuTable.putAll(skus)) }.runWith(Sink.last)
```
If I call this code in a loop with batches of a few hundreds of items, I am getting instances of `AmazonDynamoDBException`:

```
com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException: Throughput exceeds the current capacity of your table or index. DynamoDB is automatically scaling your table or index so please try again shortly. If exceptions persist, check if you have a hot key: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/bp-partition-key-design.html (Service: dynamodb; Status Code: 400; Error Code: ThrottlingException;
...
```

I would expect the exponential retry policy to take care of retrying and eventually succeeding, what am I doing wrong here?

2) when using the release `1.0.0-M10`:

```
val sc = ScanamoAlpakka(asyncDynamoDbClient, RetryPolicy.Exponential(500.milliseconds, 2.0))
val result: Future[Unit] = sc.exec(rawSkuTable.putAll(skus)) }.runWith(Sink.last).map{ result =>
      val numUnprocessed = result.map(_.getUnprocessedItems.size).sum
      if(numUnprocessed > 0) {
        throw new MyException(s"Could not write ${numUnprocessed} of ${skus.size} raw skus to DynamoDB")
      } else
        logger.info(s"Successfully wrote ${skus.size} raw skus to DynamoDB")
        Unit
}
```

In this version, at least the unprocessed items are returned, so that I can throw a custom exception (and possible retry unprocessed items only). However, sometimes my custom exception is not called and the above `AmazonDynamoDBException` is thrown instead.

I am having a hard time figuring out how to avoid write loss. Do I need to wrap my code into a custom retry-loop? If yes, what would that look like? What can I expect from the retry policies? Would the sync client API work better for such a case? The docs seem to be outdated for the versions I tried, so I'm not sure if I'm using it as intended. 

And finally, would a version of this library based on the v2 version of the AWS Java SDK make using custom retru policies obsolete, as retrying and throttling is taken care of by the AWS library iteself?

If I ditch the On-Demand mode and switch to Provided Mode, that might help, but sooner or later, with increased workload, I might find myself at the same point of getting throttled.

Sorry for asking lots of questions!

pederpansen:
…index" when using on-demand mode with Alpakka client.

I think I found out the cause for #589: The exception thrown is of the base type `AmazonDynamoDBException` instead of one of the subtypes handled in `ScanamoAlpakka.defaultRetryableCheck`.

The fix is not pretty, but easy enough.

Would you consider merging this - and including it in another release?

Resolves #589

### DIFF
--- a/alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
+++ b/alpakka/src/main/scala/org/scanamo/ScanamoAlpakka.scala
@@ -21,6 +21,7 @@ import akka.NotUsed
 import akka.stream.alpakka.dynamodb.DynamoClient
 import akka.stream.scaladsl.{ Sink, Source }
 import com.amazonaws.services.dynamodbv2.model.{
+  AmazonDynamoDBException,
   InternalServerErrorException,
   ItemCollectionSizeLimitExceededException,
   LimitExceededException,
@@ -71,7 +72,8 @@ object ScanamoAlpakka extends AlpakkaInstances {
     case _: InternalServerErrorException | _: ItemCollectionSizeLimitExceededException | _: LimitExceededException |
         _: ProvisionedThroughputExceededException | _: RequestLimitExceededException =>
       true
-    case _ => false
+    case e: AmazonDynamoDBException if e.getErrorCode.contains("ThrottlingException") => true
+    case _                                                                            => false
   }
 }
 


regiskuckaertz:
Hmm, ok that doesn't make any sense. `ThrottlingException` does not even exist in the dynamodb documentation 🤔

pederpansen:
I know, it's actually just a string in the error code of a generic exception. However, I can assure you that it does solve the issue, I tried it out in production.

regiskuckaertz:
I believe you, of course, but it is clearly documented that `BatchWriteItem` throws an `ProvisionedThroughputExceededException`. I just noticed that I had filtered out all the AWS library updates in Scala steward, as they release every day 😭 May I please ask you to test by overriding the dependency to the latest one, i.e. v1.11.716, instead of this fix? We are using 1.11.504 so I wouldn't be surprised if that was the reason...

pederpansen:
Alright, will try it out tomorrow. Might indeed be an inconsistency in the AWS SDK which has been fixed in the meantime.

pederpansen:
After bumping AWS SDK to `1.11.716`, the problem still persists. 

I dug in the aws-sdk-java Github repository for a while and found several instances of the term `ThrottlingError` in the javadocs, although such a type does not exist. The only place I found the error message mention in #589 is in [this file](https://github.com/aws/aws-sdk-java/blob/02ba4f48b9470226f104b9166efe4915632bcae7/aws-java-sdk-models/src/main/resources/models/dynamodb-2012-08-10-model.json), which is rather weird, since it is just a huge json blob.

So, as it stands, my PR is still necessary to avoid exception when using an on-demand table under heavy throughput.

regiskuckaertz:
Sorry to be hesitant here but the whole thing is confusing:

- neither versions of the SDK provide a `ThrottlingException` for Dynamo, even though other libraries do (route 53, quick sight, etc.)

- the only mention of it in the official docs is for when a table creation/update/delete operation is performed ([here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.MessagesAndCodes.http400)) which is even more mind-boggling.

- the PHP SDK [does have one](https://docs.aws.amazon.com/aws-sdk-php/v2/api/class-Aws.DynamoDb.Exception.ThrottlingException.html)

Just to clarify: what are the operations that you are performing against DynamoDB? I'm inclined to merge this change, raise an issue in the SDK repos to flag the absence of a proper exception and refactor for when they fixed it; but before that I need to understand the exact sequence of actions your program is doing. Thank you 🙏

pederpansen:
No worries, I understand you don't want to add a workaround for an issue that should be fixed upstream.

My access pattern is quite simple: I'm doing a scan followed by many (asynchronously executed) putAll batches. The throttling happens during the putAll write operations. I suspect the on-demand tables need some time to scale when confronted with a sudden spike of writes (although the docs claim the scaling happens "instantaneously").

Btw, I already raised an issue in the AWS SDK repo: 
https://github.com/aws/aws-sdk-java/issues/2222

regiskuckaertz:
Amazing! I hope they solve it soon. But this is enough for me to approve this PR. Please just add a note that this is a temporary fix with a link to the issue so I don't lose the thread.

pederpansen:
Great! The link to the issue is in the title of the PR (this PR is also cross-referenced in the issue), would that be sufficient?

pederpansen:
Done. Thank you as well!

WtrLmmrs:
So we ran into this problem as well. Actually, it's hard to reproduce but I have seen it twice. It seems this fix is 'official' according to the documentation for ThrottlingException here https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html

However maybe a better (non-temp) fix could be to use RetryUtils.isThrottlingException(SdkException)? https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/retry/RetryUtils.html

(I'd make a PR but I have no way of reproducing this (yet))

PR #627, opened 2020-03-04
Drop 2.11

regiskuckaertz:
Both cats and magnolia are doing it, but this is an overall move in the library ecosystem. I'm not particularly for or against, in the sense that I'd like to remove at least the cats dependency in the future – we can do without magnolia/shapeless starting from Scala 3.


regiskuckaertz:
Excellent, if I can get this approved then I’ll merge 💪

PR #629, opened 2020-03-05
Upgrade ZIO and interop-cats

wi101:


renancaju:
Any chances of another release so ZIO users can get this change? Thanks!

PR #650, opened 2020-03-14
Introduce transactional (update & delete) methods

rmckirby:
Introduces the ability to action transactional updates/deletes.

### DIFF
--- a/alpakka/src/test/scala/org/scanamo/ScanamoAlpakkaSpec.scala
+++ b/alpakka/src/test/scala/org/scanamo/ScanamoAlpakkaSpec.scala
@@ -600,7 +600,7 @@ class ScanamoAlpakkaSpec extends FunSpecLike with BeforeAndAfterAll with Matcher
     }
   }
 
-  it("transact table write items") {
+  it("transact table write (put) items") {
     LocalDynamoDB.usingRandomTable(client)("name" -> S) { t =>
       val itemsTable = Table[Item](t)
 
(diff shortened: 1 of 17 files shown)


scanamo/src/main/scala/org/scanamo/request/ScanamoRequest.scala
```diff
@@ -72,3 +73,28 @@ case class RequestCondition(
   attributeNames: Map[String, String],
   dynamoValues: Option[DynamoObject]
 )
+
+case class TransactPutItem(
+  tableName: String,
+  item: DynamoValue,
+  condition: Option[RequestCondition]
```
rmckirby: Not yet used, but I've left them here should future work want to implement adding the condition expressions

alpakka/src/test/scala/org/scanamo/ScanamoAlpakkaSpec.scala
```diff
@@ -641,4 +641,97 @@ class ScanamoAlpakkaSpec extends FunSpecLike with BeforeAndAfterAll with Matcher
       }
     }
   }
+
+  it("transact table write (update) items") {
```
rmckirby: There's a fair but of duplication between the different specs here, not sure if we see some value in reducing that. For now I've just followed the existing style which seems ok.

regiskuckaertz: Agreed. There is a way around that using some form of `ZManaged` or similar and feeding tables with a random sample of data, then all these tests will be reduced to "laws". One day 😩

rmckirby:
> ❤️ this is amazing work! Thanks a lot for your first contribution, this makes a very useful feature of DynamoDB safe and easy to work with 💪 For heterogeneous transactions, we will need a new construct that can work across tables and accept any kind of CUD operations.

Thanks @regiskuckaertz, I'm happy to contribute!
Do you have a timeframe on when we can expect a new release with these changes in?

PR #802, opened 2020-07-30
Build speedup

regiskuckaertz:
Also updates Scala versions


regiskuckaertz:
Happy with that.

Before

Screenshot 2020-07-30 at 08 29 32

After

Screenshot 2020-07-30 at 08 29 39

PR #875, opened 2020-10-28
Add DynamoFormat for DynamoValue

benjumanji:
It is useful for debugging to be able to magic up a table that pulls the
raw data down without trying to decode it.

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/DynamoValue.scala
+++ b/scanamo/src/main/scala/org/scanamo/DynamoValue.scala
@@ -358,4 +358,9 @@ object DynamoValue {
   def fromDynamoArray(xs: DynamoArray): DynamoValue = DynArray(xs)
 
   private[scanamo] def unsafeFromNumber(n: String): DynamoValue = DynNum(n)
+
+  implicit object dynamoValueFormat extends DynamoFormat[DynamoValue] {
+    def read(av: DynamoValue): Either[DynamoReadError, DynamoValue] = Right(av)
+    def write(av: DynamoValue): DynamoValue = av
+  }
 }


regiskuckaertz (approved):
Good catch! Thank you.

priyankadhiman0803:
No, it'snot the same. You see, that is trait and is being extended by two classes. I think i would have to create a new ObjectFormat instead of Dynamo format.I tried this

- [ ] implicit val dynamoFormatRules : DynamoFormat[UserDuplicateRules] = new ObjectFormat[UserDuplicateRules] {
- [ ]     override def readObject(o: DynamoObject): Either[DynamoReadError, UserDuplicateRules] = {
- [ ]      val rules = o.get[String]("rule")
- [ ]      val keep = o.get[DuplicateDetails]("keep")
- [ ]       (rules, keep) match {
- [ ]         case (Right("NoUserLoggedIn"), Right(keepValue)) => Right(NoUserLogged(keepValue))
- [ ]         case (Right("OnlyOneuserLogged"), Right(keepValue) ) => Right(OnlyOneuserLogged(keepValue))
- [ ]         case _ => Left(TypeCoercionError(new Exception(" invalid value for user response")))
- [ ]       }
- [ ]     }
- [ ] 
- [ ]     override def writeObject(t: UserDuplicateRules) : DynamoObject = {
- [ ]       t match {
- [ ]         case noUserLogged: NoUserLogged =>  DynamoObject(Map[String, DynamoValue]("rule" -> DynamoValue.fromString("NoUserLoggedIn"),
- [ ]           "keep" -> DynamoValue.fromDynamoObject(DuplicateDetails.createDynamoObject(noUserLogged.keep))

- [ ]          ))
- [ ] 
- [ ]         case onlyuserLogged: OnlyOneuserLogged => DynamoObject(Map[String, DynamoValue]("rule" -> DynamoValue.fromString("OnlyOneuserLogged"),
- [ ]           "keep" -> DynamoValue.fromDynamoObject(DuplicateDetails.createDynamoObject(onlyuserLogged.keep))

- [ ]         ))
- [ ]       }
- [ ]     }
- [ ]   }


write is working fine but not the read . plus i am not sure how to read the List of same object . I mean how to convert a List of Dynamo Object into DynamoObject . There is nothing mentioned in the documentation .

PR #889, opened 2020-11-17
Fix bug in deleteAndReturn

### LINKED ISSUE #879: deleteAndReturn always returns None
The following code always returns `None` even though the item to delete exists in DynamoDB. 

```
val op = for {
      result <- apiKeysTable.deleteAndReturn(DeleteReturn.OldValue)("enc_key" === hash)
    } yield result

    client.exec(op).map {
      case Some(value) => {
        value match {
          case Right(_) => true
          case Left(_)  => false
        }
      }

      case None => false
    }
```

Similar code properly returns `false` when the item is not found and `true` when a delete operation was performed.
```
val op = for {
      result <- apiKeysTable.when(attributeExists("enc_key")).delete("enc_key" === key)
    } yield result

    client.exec(op).map {
      case Right(_) => true
      case Left(_)  => false
    }
```
Scanamo version: `1.0-M13`
Scala version: `2.12.11`

davidfurey:
DeleteReturn was being ignored, this adds it to the dynamodb request to resolve #879

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/ops/package.scala
+++ b/scanamo/src/main/scala/org/scanamo/ops/package.scala
@@ -125,7 +125,11 @@ package object ops {
     }
 
     def delete(req: ScanamoDeleteRequest): DeleteItemRequest = {
-      val request = DeleteItemRequest.builder.tableName(req.tableName).key(req.key.toJavaMap)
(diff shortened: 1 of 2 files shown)


regiskuckaertz (approved):
Thank you!

PR #974, opened 2021-02-09
Unwrap CompletionException

joroKr21:
Thanks a lot Java!
Why doesn't `toScala` do this? :cry:


regiskuckaertz (approved):
Beautiful! Thanks a lot for your contribution, this is looking great 🕺

PR #977, opened 2021-02-10
Remove references to com.gu

davidfurey:
Scanamo has moved to the org.scanamo organisation.


regiskuckaertz (approved):
Dammit, I can't believe they slipped through the cracks.

Issue #987, opened 2021-02-17
scan + limit query hangs forever


simpadjo:
Hi, thanks for the library!

I faced the following issue:
I do full scan. The query matches one entry in the table. 
Then I add `limit(2)` restriction and it hangs forever.

`case class Row(id: String, field: String)

object ScanBug extends App {
  val cl: DynamoDbAsyncClient = ???
  
  val zioCl = ScanamoZio(cl)
  val table = Table[Row]("table")
  
  //this query matches 1 entry
  val q = table.filter(("field" === "some-value")).limit(2).scan()
  val res = zio.Runtime.default.unsafeRunTask( zioCl.exec(q))
  println(res)
}`

It is 100% reproducible in production, haven't yet reproduced it with local Dynamo.
If I remove `limit`, change `limit(2)` to `limit(1000)` or change `scan()` to `scan0` - everything is OK.

I suspect that the problem has something to do with span response pagination: 
`table.filter(("field" === "some-value")).limit(2).scan0` successfully returns 0 results with a continuation token.

simpadjo:
(This is the first day I'm working with DynamoDb, so might be talking stupid things)

I don't understand this logic:

https://github.com/scanamo/scanamo/blob/master/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala#L44

```
def stream[T: DynamoFormat](req: Req): ScanamoOps[(List[Either[DynamoReadError, T]], Option[DynamoObject])] = {
    def streamMore(req: Req): ScanamoOps[(List[Either[DynamoReadError, T]], Option[DynamoObject])] =
      for {
        res <- exec(req)
        results = items(res).map(ScanamoFree.read[T])
        newLimit = limit(req).map(_ - results.length)
        lastKey = lastEvaluatedKey(res).filterNot(_.isEmpty)
        result <-
          lastKey
            .filterNot(_ => newLimit.exists(_ <= 0))
            .foldLeft(
```

`limit` in AWS api corresponds to the number of source entries to scan rather than the number of target entries to return. 
`newLimit = limit(req).map(_ - results.length)`  - this looks like an attempt to collect up to `limit` result entries.

Now I understand why my query 'hangs': if a table contains ~10k entries, it takes 5k queries to traverse it.
I tested it from outside of AWS network so it seemed like hanging. 
This also explains why increasing `limit` helps and why I can't reproduce it with local Dynamo instance.

regiskuckaertz:
[...]

PR #996, opened 2021-02-22
Reduce limit based on scannedCount not returned results

davidfurey:
Dynamodb applies the `limit` parameter before the FilterExpression is evaluated, so it represents the maximum number of items that will be read, the actual number of items returned may be lower.

```
A single Scan operation reads up to the maximum number of items set (if using the Limit parameter) 
or a maximum of 1 MB of data and then apply any filtering to the results using FilterExpression. If 
LastEvaluatedKey is present in the response, you need to paginate the result set. For more 
information, see Paginating the Results in the Amazon DynamoDB Developer Guide.
```
(https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html)

When paginating with a limit parameter, we need to reduce the limit with each request so that the total number of items read corresponds to the desired limit, rather than corresponding to `totalPages * limit`

We were erroneously subtracting the number of items returned from the `limit`, but we should subtract the number of items read (pre-filtering).  Conveniently the Dynamodb API includes this as the `ScannedCount` in its response.  This PR switches to using `ScannedCount` for the subtraction.

Related to #987

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
+++ b/scanamo/src/main/scala/org/scanamo/DynamoResultStream.scala
@@ -29,6 +29,7 @@ private[scanamo] trait DynamoResultStream[Req, Res] {
   def startKey(req: Req): Option[DynamoObject]
   def items(res: Res): List[DynamoObject]
   def lastEvaluatedKey(res: Res): Option[DynamoObject]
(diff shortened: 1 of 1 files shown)


regiskuckaertz (approved):
Good catch!

PR #1023, opened 2021-03-20
Fix a bug in `ScanResponseStream`

kailuowang:
I noticed a bug with `scanPaginatedM` namely dynamo returns an empty object as last key when the result size is less than the limit, i.e. there are no more pages.  However `ScanResponseStream` doesn't recognize this and still try to use the empty object as the next start key causing a 
```
software.amazon.awssdk.services.dynamodb.model.DynamoDbException: Exclusive Start Key must have same size as table's key schema
```
This PR changed one of the tests to expose this error and fixed the bug by filtering out empty last key.


regiskuckaertz (approved):
Yikes! Thank you @kailuowang 👍

Issue #1083, opened 2021-05-04
Support different field naming conventions in `DynamoFormat`


VladKopanev:
I think it would be great to be able to support something similar to 
https://circe.github.io/circe/codecs/custom-codecs.html#custom-key-mappings-via-annotations
but for `DynamoFormat`.
I have a case where fields in Dynamo are named in UpperCamelCase, but I want to stick with Scala fields naming convention in my model case classes.

PR #1128, opened 2021-05-29
#1083 Add field naming modes

VladKopanev:
This change adds a new `FieldNamingMode` trait that should be mixed into Derivation trait implementation (self-type requirement).  
This trait has `transformName `method that is responsible for transforming the case class parameters names.
The default implementation of this new trait is `NoOpFieldNamingMode` that is basically an identity function.
For specific naming conventions there are new objects in the packages `org.scanamo.generic.auto` and `org.scanamo.generic.semiauto`, new objects are:
* `kebabCase`
* `pascalCase`
* `snakeCase`

So, if the users want Scanamo to generate DynamoFormats for classes that have a related table with snake case fields in it, they just need to add this import: `import org.scanamo.generic.auto.snakeCase._`.

Alternatively, users can create their own implementations of `FieldNamingMode ` and mix it with `Derivation` trait.

### DIFF
--- a/scanamo/src/main/scala-2.x/org/scanamo/generic/FieldNamingMode.scala
+++ b/scanamo/src/main/scala-2.x/org/scanamo/generic/FieldNamingMode.scala
@@ -31,5 +31,3 @@ trait KebabCaseFieldNamingMode extends FieldNamingMode {
       .replaceAll(pattern.pattern(), "$1-$2")
       .toLowerCase()
 }
-
-
--- a/scanamo/src/main/scala-2.x/org/scanamo/generic/FieldNamingMode.scala
+++ b/scanamo/src/main/scala-2.x/org/scanamo/generic/FieldNamingMode.scala
@@ -8,12 +8,12 @@ trait FieldNamingMode {
 
 trait NoOpFieldNamingMode extends FieldNamingMode {
 
-  @inline final override def transformName(fieldName: String): String = fieldName
+  final override def transformName(fieldName: String): String = fieldName
 }
 
 trait PascalCaseFieldNamingMode extends FieldNamingMode {
 
-  @inline final override def transformName(fieldName: String): String = fieldName.capitalize
+  final override def transformName(fieldName: String): String = fieldName.capitalize
 }
 
 trait SnakeCaseFieldNamingMode extends FieldNamingMode {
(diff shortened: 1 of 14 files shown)


scanamo/src/main/scala-2.x/org/scanamo/generic/FieldNamingMode.scala
```diff
@@ -0,0 +1,33 @@
+package org.scanamo.generic
+
+import java.util.regex.Pattern
+
+trait FieldNamingMode {
```
regiskuckaertz, line 5: Let's mark these as package-private

scanamo/src/main/scala-2.x/org/scanamo/generic/auto/kebabCase.scala
```diff
@@ -0,0 +1,8 @@
+package org.scanamo.generic.auto
+
+import org.scanamo.DynamoFormat
+import org.scanamo.generic.{ AutoDerivation, Exported, KebabCaseFieldNamingMode }
+
+object kebabCase extends AutoDerivation with KebabCaseFieldNamingMode {
+  implicit final def genericDerivedFormat[A]: Exported[DynamoFormat[A]] = macro materializeImpl[A]
```
regiskuckaertz, line 7: You will need to `import scala.language.experimental.macros` everywhere.

regiskuckaertz (changes requested):
Nice stuff! The only criticism to using the cake pattern here is that new configuration options multiply the mixins composition. Instead we could make that configuration an implicit parameter in the case of auto-derivation (by default NoOp, which could be overwritten by an implicit in local scope) and an explicit parameter in the case of semiauto-derivation (which would have a default NoOp value)

VladKopanev:
> Nice stuff! The only criticism to using the cake pattern here is that new configuration options multiply the mixins composition. Instead we could make that configuration an implicit parameter in the case of auto-derivation (by default NoOp, which could be overwritten by an implicit in local scope) and an explicit parameter in the case of semiauto-derivation (which would have a default NoOp value)

Hmmm, interesting idea, but I'm not sure I can pass implicit or whatever parameter through magnolia to the derivation code. Do you have any ideas on how can it be done?

PR #1156, opened 2021-07-06
Fix 404ing docs links

kittsville:
Internal docs links are pointing to the source filename rather than the generated page


philwills:
Thanks for fixing this, sorry it's taken so long to get merged.

kittsville:
No worries at all. These are trying times, look after yourself 💜

PR #1163, opened 2021-07-09
use Scaladex badge on ReadMe to show Scala version support

rtyley:
This badge summarises which versions of Scala are supported by Scanamo (and what the latest Scanamo version is for each of those Scala versions):

[scanamo Scala version support](https://index.scala-lang.org/scanamo/scanamo/scanamo)

More details on the badge format: https://github.com/scalacenter/scaladex/pull/660


regiskuckaertz (commented):
The homepage needs to be updated too here: https://github.com/scanamo/scanamo/blob/master/docs/docs/index.md

rtyley:
> The homepage needs to be updated too here: https://github.com/scanamo/scanamo/blob/master/docs/docs/index.md

Ah thanks, I missed that! I've amended the commit and force-pushed, should cover both places now.

regiskuckaertz (approved):

rtyley:
> The homepage needs to be updated too here: https://github.com/scanamo/scanamo/blob/master/docs/docs/index.md

The homepage at https://www.scanamo.org/ doesn't seem to have updated (still shows the maven central badge), is that an automated process?

image

regiskuckaertz:
Hmm yes I think I never got around to adding it to the build and was doing it manually. Iirc it should be just a single sbt instruction now; with the only caveat that the CNAME file always seems to disappear for some reason and I had to recreate it.

PR #1167, opened 2021-07-11
Upgrade to sbt v1.5

rtyley:
The changes required by this update are:

* Removing use of the [deprecated sbt-0.13-style-DSL **` in `** syntax](https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#Migrating+to+slash+syntax)
* Tweaks for sbt v1.5 eviction errors (see sbt/sbt#6221):
  * Upgrade `cats-effect` to v3.1.1 - see https://github.com/scanamo/scanamo/pull/1166, which should be merged before this PR
  * Set `evictionErrorLevel := Level.Info`, because while we can fix `cats-effect`, we can't currently fix `akka-http` - the current latest version of `akka-http` suffers from a semver issue with `scala-java8-compat` - see akka/akka#30375 :
  ```
  [info] 	* org.scala-lang.modules:scala-java8-compat_2.12:1.0.0 (early-semver) is selected over 0.8.0
  [info] 	    +- org.scanamo:scanamo_2.12:1.0.0-M15+99-399e98af+20210710-2246-SNAPSHOT (depends on 1.0.0)
  [info] 	    +- com.typesafe.akka:akka-actor_2.12:2.5.31           (depends on 0.8.0)
  ```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ scalaVersion in ThisBuild := "2.12.14"
 crossScalaVersions in ThisBuild := Seq("2.12.14", "2.13.6")
 
 val catsVersion = "2.6.1"
-val catsEffectVersion = "2.5.1"
+val catsEffectVersion = "3.1.1"
 val zioVersion = "1.0.9"
 
 lazy val stdOptions = Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,5 @@
-scalaVersion in ThisBuild := "2.12.14"
-crossScalaVersions in ThisBuild := Seq("2.12.14", "2.13.6")
+ThisBuild / scalaVersion := "2.12.14"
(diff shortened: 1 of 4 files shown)


regiskuckaertz (approved):

build.sbt
```diff
@@ -227,9 +227,10 @@ lazy val alpakka = (project in file("alpakka"))
       "org.scalatest"      %% "scalatest"                    % "3.2.9"  % Test,
       "org.scalacheck"     %% "scalacheck"                   % "1.15.4" % Test
     ),
-    fork in Test := true,
+    evictionErrorLevel := Level.Info, // until Akka 2.6.16 released - see https://github.com/akka/akka/pull/30375
```
rtyley, line 230: Needed to do this on the `docs` subproject too! See #1176

PR #1174, opened 2021-07-16
Remove instance-method version of `coercedXmap`

rtyley:
The instance method version of `coercedXmap` on the `DynamoFormat` trait delegates to the static `DynamoFormat.coercedXmap` method - but that requires an implicit `ClassTag` for `T`, the `Throwable` type that it's allowed to coerce:

https://github.com/scanamo/scanamo/blob/4f79e47e1387c98a48120114050853d86c8df0c5/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala#L122-L126

The instance method version of `coercedXmap` can't know what that `Throwable` type is, so far as I can see, with it's current signature?

https://github.com/scanamo/scanamo/blob/4f79e47e1387c98a48120114050853d86c8df0c5/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala#L59-L60

The Scala 3 compiler spots this, and errors with a `No ClassTag available` message:

```
[error] -- Error: /home/roberto/development/scanamo/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala:60:58
[error] 60 |    DynamoFormat.coercedXmap(read, write)(this, implicitly)
[error]    |                                                          ^
[error]    | No ClassTag available for T
[error]    |
[error]    | where:    T is a type variable with constraint >: Null and <: Throwable
```

...the Scala 2.13 compiler doesn't seem to have a problem with it, but probably should?!

As the `coercedXmap` instance method does not actually allow users to specify what subtype of `Throwable` they want to coerce, I suspect it's unused. Indeed, all the tests continue to compile with it removed. Consequently, I think it makes sense to delete it.

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
+++ b/scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
@@ -56,8 +56,6 @@ trait DynamoFormat[T] {
 
   def iso[U](r: T => U, w: U => T): DynamoFormat[U] = DynamoFormat.iso(r, w)(this)
   def xmap[U](r: T => Either[DynamoReadError, U], w: U => T): DynamoFormat[U] = DynamoFormat.xmap(r, w)(this)
-  def coercedXmap[U](read: T => U, write: U => T): DynamoFormat[U] =
-    DynamoFormat.coercedXmap(read, write)(this, implicitly)
 }
 
 object DynamoFormat extends PlatformSpecificFormat {


scanamo/src/main/scala/org/scanamo/DynamoFormat.scala
```diff
@@ -56,8 +56,6 @@ trait DynamoFormat[T] {
 
   def iso[U](r: T => U, w: U => T): DynamoFormat[U] = DynamoFormat.iso(r, w)(this)
   def xmap[U](r: T => Either[DynamoReadError, U], w: U => T): DynamoFormat[U] = DynamoFormat.xmap(r, w)(this)
-  def coercedXmap[U](read: T => U, write: U => T): DynamoFormat[U] =
```
regiskuckaertz, line 59: Yeah, the type is a constraint on [`catchOnly`](https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/core/src/main/scala/cats/syntax/either.scala#L365). I'm not so sure removing is the right way to go, as people may be using it. I actually don't like the companion methods, it is way more idiomatic to write:
```scala
DynamoFormat[Int].coercedXmap(_.toString, _.toInt)
```
True, scanamo does not have good story for manually-written instances, but it could with a improvements.

This should please the scala 3 compiler
```
DynamoFormat.coercedXmap(read, write)(this, implicitly[ClassTag[Throwable]])
```

After all, `Throwable >: Null <: Throwable` and functions are by nature impure in Scala. In Scala 3 we may be able to change the type to:

```scala
def coercedXmap[Ex >: Null <: Throwable: ClassTag]
  ( read: T => U canThrow Ex
  , write: U => T canThrow Ex
  ): DynamoFormat[U]
```
... [one day perhaps](https://github.com/lampepfl/dotty/blob/b65958f139a0a818d36ee04468ada93b8fb472ac/docs/docs/reference/experimental/canthrow.md).

rtyley: > I'm not so sure removing is the right way to go, as people may be using it.

This is true, it's hard to know. I did just now do some searches on GitHub, to see if I could find any examples of it's use. I checked:

* The 40 occurrences of `coercedXmap` in [the public & private repos of the Guardian](https://github.com/search?q=org%3Aguardian+coercedXmap&type=Code)
* The 31 occurrences of `coercedXmap` in [public non-Guardian repos](https://github.com/search?q=coercedXmap+-org%3Aguardian&type=Code)

...**every single one of those occurrences was using the static `DynamoFormat.coercedXmap` form** - none of them using the instance method. Of course, there may well be lots of code in private repositories that I can't see that _is_ using the instance version!

> This should please the scala 3 compiler
> 
> ```
> DynamoFormat.coercedXmap(read, write)(this, implicitly[ClassTag[Throwable]])
> ```

I've given this a try out, and it does indeed please the Scala 3 compiler, thank you! However, although I agree that the idiomatic form of `DynamoFormat[Int].coercedXmap(_.toString, _.toInt)` is potentially nice, the fact that it seems currently unused, and perhaps catches a lot more exceptions than users would expect (eg `java.lang.OutOfMemoryError`) I think a better solution probably _would_ be to remove the method.

My focus for the moment is getting Scala 3 working though, so if you feel it's better to do the `implicitly` fix, I'll raise a new PR for that.

regiskuckaertz (approved):
I think adding it later with a better signature makes sense.

PR #1188, opened 2021-08-15
Scala 3 prep: Upgrade to magnolia 1.0

rtyley:
Only the most recent releases of the Magnolia library are available for both Scala 2 and 3 - this is since the project maintainership moved to the https://github.com/softwaremill organisation. Note that as part of this move, there's an updated group id, and package name of `magnolia1` - some details about that here: https://github.com/softwaremill/magnolia/issues/327

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -131,9 +131,9 @@ lazy val scanamo = (project in file("scanamo"))
   .settings(
     libraryDependencies ++= Seq(
       awsDynamoDB,
-      "org.scala-lang.modules" %% "scala-java8-compat" % "1.0.0",
-      "org.typelevel"          %% "cats-free"          % catsVersion,
-      "com.propensive"         %% "magnolia"           % "0.12.7",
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -131,9 +131,9 @@ lazy val scanamo = (project in file("scanamo"))
   .settings(
     libraryDependencies ++= Seq(
       awsDynamoDB,
-      "org.scala-lang.modules" %% "scala-java8-compat" % "1.0.0",
-      "org.typelevel"          %% "cats-free"          % catsVersion,
-      "com.propensive"         %% "magnolia"           % "0.12.7",
+      "org.scala-lang.modules"       %% "scala-java8-compat" % "1.0.0",
+      "org.typelevel"                %% "cats-free"          % catsVersion,
+      "com.softwaremill.magnolia1_2" %% "magnolia"           % "1.0.0-M5",
```
regiskuckaertz, line 136: This is where having version-specific dependencies will become a necessity.

regiskuckaertz (approved):

PR #1324, opened 2022-01-10
Improved index paging

chrislewis:
This PR aims to improve the situation around paging over indexes by adding a couple features (loosely defined), as well as a change to `SecondaryIndex#from`:

1. Enables access to the underlying response from DynamoDB when querying or scanning an index. This is simply the addition of `scan0` and `query0` to `SecondaryIndex`.
2. Enables the construction of index keys having 3-4 properties using the established syntax.
3. Changes the signature of `SecondaryIndex#from` to constrain it against `IndexKey`.

The last one is potentially controversial so here is the reasoning (and of course if someone has a clearer vision of a better approach, I'd love to hear it and help if I can - new here):

Previously,  `SecondaryIndex#from` had the same constraints as `Table#from` (`UniqueKeyCondition`), however this is insufficient given the keys for indexes can and often do differ in their structure. That is, it is common to have a secondary index with a different hash key, sort key, or both. When paging over an index, the last evaluated key must contain all properties comprising both the key of the index and that of its parent table. As far as I can tell the only way to construct such a key would be to provide another instance of `UniqueKeyCondition` over some type with additional properties. However, that would then make it possible to pass invalid/impossible keys to `Table#from`, which will only ever have 1 or 2 properties.

This change breaks that, which could break existing uses of `SecondaryIndex#from`. However as they have different requirements in reality, this seems appropriate. Further, with the exception of any code that has implemented its own workaround by providing (potentially dangerous) instances of `UniqueKeyCondition`, the support for the established syntax should keep those source compatible.

Even with workarounds in terms of `UniqueKeyCondition`, I still don't see that it's possible to get the underlying response when interacting with an index. This means that one would have to compute the last evaluated key (which is possible), but still couldn't know if there was actually a subsequent page.

Again if someone has a better vision/approach, I'd be glad to help or just see it merged (this is a blocking issue for my team's adoption)! Thanks to all committers + maintainers for a great library ❤️

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/SecondaryIndex.scala
+++ b/scanamo/src/main/scala/org/scanamo/SecondaryIndex.scala
@@ -16,11 +16,12 @@
 
 package org.scanamo
 
-import cats.{ Monad, MonoidK }
-import org.scanamo.DynamoResultStream.{ QueryResponseStream, ScanResponseStream }
-import org.scanamo.ops.{ ScanamoOps, ScanamoOpsT }
-import org.scanamo.query.{ Condition, ConditionExpression, Query, UniqueKey, UniqueKeyCondition }
-import org.scanamo.request.{ ScanamoQueryOptions, ScanamoQueryRequest, ScanamoScanRequest }
+import cats.{Monad, MonoidK}
+import org.scanamo.DynamoResultStream.{QueryResponseStream, ScanResponseStream}
+import org.scanamo.ops.{ScanamoOps, ScanamoOpsT}
+import org.scanamo.query.{Condition, ConditionExpression, IndexKey, Query}
+import org.scanamo.request.{ScanamoQueryOptions, ScanamoQueryRequest, ScanamoScanRequest}
+import software.amazon.awssdk.services.dynamodb.model.{QueryResponse, ScanResponse}
 
 /** Represents a secondary index on a DynamoDB table.
   *
@@ -49,9 +50,17 @@ sealed abstract class SecondaryIndex[V] {
     */
   def scanPaginatedM[M[_]: Monad: MonoidK](pageSize: Int): ScanamoOpsT[M, List[Either[DynamoReadError, V]]]
 
+  /** Scans the table and returns the raw DynamoDB result.
+    */
+  def scan0: ScanamoOps[ScanResponse]
+
   /** Run a query against keys in a secondary index
     */
   def query(query: Query[_]): ScanamoOps[List[Either[DynamoReadError, V]]]
+
+  /** Queries the table and returns the raw DynamoDB result.
+    */
+  def query0(query: Query[_]): ScanamoOps[QueryResponse]
 
   /** Performs a query with the ability to introduce effects into the computation. This is
     * useful for huge tables when you don't want to load the whole of it in memory, but
@@ -85,7 +94,7 @@ sealed abstract class SecondaryIndex[V] {
 
   def descending: SecondaryIndex[V]
 
-  def from[K: UniqueKeyCondition](key: UniqueKey[K]): SecondaryIndex[V]
+  def from[T, K](key: T)(implicit K: IndexKey.Aux[T, K]): SecondaryIndex[V]
 }
 
 private[scanamo] case class SecondaryIndexWithOptions[V: DynamoFormat](
@@ -94,8 +103,8 @@ private[scanamo] case class SecondaryIndexWithOptions[V: DynamoFormat](
   queryOptions: ScanamoQueryOptions
 ) extends SecondaryIndex[V] {
   def limit(n: Int): SecondaryIndexWithOptions[V] = copy(queryOptions = queryOptions.copy(limit = Some(n)))
-  def from[K: UniqueKeyCondition](key: UniqueKey[K]) =
-    copy(queryOptions = queryOptions.copy(exclusiveStartKey = Some(key.toDynamoObject)))
+  def from[T, K](key: T)(implicit K: IndexKey.Aux[T, K]) =
+    copy(queryOptions = queryOptions.copy(exclusiveStartKey = Some(IndexKey.toDynamoObject(key))))
   def filter[C: ConditionExpression](condition: C) =
     SecondaryIndexWithOptions[V](tableName, indexName, ScanamoQueryOptions.default).filter(Condition(condition))
   def filter[T](c: Condition[T]): SecondaryIndexWithOptions[V] =
@@ -105,8 +114,12 @@ private[scanamo] case class SecondaryIndexWithOptions[V: DynamoFormat](
   def scan() = ScanResponseStream.stream[V](ScanamoScanRequest(tableName, Some(indexName), queryOptions)).map(_._1)
   def scanPaginatedM[M[_]: Monad: MonoidK](pageSize: Int) =
     ScanResponseStream.streamTo[M, V](ScanamoScanRequest(tableName, Some(indexName), queryOptions), pageSize)
+  def scan0: ScanamoOps[ScanResponse] =
+    ScanamoOps.scan(ScanamoScanRequest(tableName, Some(indexName), queryOptions))
   def query(query: Query[_]) =
     QueryResponseStream.stream[V](ScanamoQueryRequest(tableName, Some(indexName), query, queryOptions)).map(_._1)
+  def query0(query: Query[_]): ScanamoOps[QueryResponse] =
+    ScanamoOps.query(ScanamoQueryRequest(tableName, Some(indexName), query, queryOptions))
   def queryPaginatedM[M[_]: Monad: MonoidK](query: Query[_], pageSize: Int) =
     QueryResponseStream.streamTo[M, V](ScanamoQueryRequest(tableName, Some(indexName), query, queryOptions), pageSize)
 }
--- a/scanamo/src/main/scala/org/scanamo/SecondaryIndex.scala
+++ b/scanamo/src/main/scala/org/scanamo/SecondaryIndex.scala
@@ -16,12 +16,12 @@
 
 package org.scanamo
 
-import cats.{Monad, MonoidK}
-import org.scanamo.DynamoResultStream.{QueryResponseStream, ScanResponseStream}
-import org.scanamo.ops.{ScanamoOps, ScanamoOpsT}
-import org.scanamo.query.{Condition, ConditionExpression, IndexKey, Query}
-import org.scanamo.request.{ScanamoQueryOptions, ScanamoQueryRequest, ScanamoScanRequest}
-import software.amazon.awssdk.services.dynamodb.model.{QueryResponse, ScanResponse}
+import cats.{ Monad, MonoidK }
+import org.scanamo.DynamoResultStream.{ QueryResponseStream, ScanResponseStream }
+import org.scanamo.ops.{ ScanamoOps, ScanamoOpsT }
+import org.scanamo.query.{ Condition, ConditionExpression, IndexKey, Query }
+import org.scanamo.request.{ ScanamoQueryOptions, ScanamoQueryRequest, ScanamoScanRequest }
+import software.amazon.awssdk.services.dynamodb.model.{ QueryResponse, ScanResponse }
 
 /** Represents a secondary index on a DynamoDB table.
   *
--- a/scanamo/src/main/scala/org/scanamo/package.scala
+++ b/scanamo/src/main/scala/org/scanamo/package.scala
@@ -18,6 +18,7 @@ package org
 
 import org.scanamo.query._
 import org.scanamo.update._
+
 import scala.language.implicitConversions
 
 package object scanamo {
@@ -77,6 +78,14 @@ package object scanamo {
 
     implicit class OrConditionExpression[X: ConditionExpression](x: X) {
       def or[Y: ConditionExpression](y: Y): OrCondition[X, Y] = OrCondition(x, y)
+    }
+
+    /** Syntax for `AndEqualsCondition` instances over a pair of `KeyEquals` with an `and` method to support
+      * construction of index keys of 3 properties.
+      */
+    implicit class AndEqualsConditionOps[H, S](ae: AndEqualsCondition[KeyEquals[H], KeyEquals[S]]) {
+      def and[I](ke: KeyEquals[I]): IndexKey3[KeyEquals[H], KeyEquals[S], KeyEquals[I]] =
+        IndexKey3(ae.hashEquality, ae.rangeEquality, ke)
     }
 
     @deprecated("use uncurried `set(attr, value)` syntax", "1.0")
--- a/scanamo/src/main/scala/org/scanamo/query/IndexKey.scala
+++ b/scanamo/src/main/scala/org/scanamo/query/IndexKey.scala
@@ -1,0 +1,126 @@
+package org.scanamo.query
+
+import cats.implicits._
+import org.scanamo.DynamoObject
+
+/** A lawless typeclass functionally identical to [[UniqueKeyCondition]] but used to constrain a different set values
+  * to be used as index keys.
+  */
+trait IndexKey[T] {
+  type K
+  def toDynamoObject(t: T): DynamoObject
+  def fromDynamoObject(key: K, dvs: DynamoObject): Option[T]
+  def key(t: T): K
+}
+
+/** An index key with 3 properties to support paginating indexes whose key structure has a property that does not
+  * overlap with that of the parent table.
+  */
+final case class IndexKey3[A, B, C](a: A, b: B, c: C) {
+  def and[D](d: D): IndexKey4[A, B, C, D] = IndexKey4(a, b, c, d)
+}
+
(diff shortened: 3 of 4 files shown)


kailuowang:
First, quick question is there a reason for the signature to be 
```scala
def from[T, K](key: T)(implicit K: IndexKey.Aux[T, K])
```
instead of 
```scala
def from[T](key: T)(implicit K: IndexKey[T])
```
but that being said the original signature is a bit redundant too. 
```scala
  def from[K: UniqueKeyCondition](key: UniqueKey[K])
```
The implicit conversion from `K` to `UniqueKey[K]` isn't necessary and makes the API a bit less clear. I would prefer 
```scala
  def from[K: UniqueKeyCondition](key: K)
```
I think that change will be source-compatible. 

If we go with this signature, there is much difference between 
```scala
  def from[K: UniqueKeyCondition](key: K)
```
and 
```scala
  def from[K: IndexKey](key: K)
```
I think we can achieve what is provided by `IndexKey` with `UniqueKeyCondition` by providing a UniqueKeyCondition instance for `(KeyEquals, KeyEquals)` , `(KeyEquals, KeyEquals, KeyEquals)`  and so on and force.

But I don't think we have to provide a bunch of instance methods for `UniqueKeyCondition` for different tuple sizes like in `IndexKey`? 
I think a more composable way to provide `UniqueKeyCondition` for all kinds of shapes will be something like this
```Scala
 implicit def uniqueKeyConditionTuple[A, B](implicit
      A: UniqueKeyCondition[A],
      B: UniqueKeyCondition[B]
  ): UniqueKeyCondition[(A, B)] =
    new UniqueKeyCondition[(A, B)] {
      type K = (A.K, B.K)
      final def toDynamoObject(t: (A, B)): DynamoObject =
        A.toDynamoObject(t._1) <> B.toDynamoObject(t._2)
      final def fromDynamoObject(
          key: K,
          dvs: DynamoObject
      ): Option[(A, B)] =
        (A.fromDynamoObject(key._1, dvs), B.fromDynamoObject(key._2, dvs))
          .mapN((_, _))
      final def key(t: (A, B)): K = (A.key(t._1)), B.key(t._2))
    }

```
Then you can use `from` like for primary key id and secondary composite key "name" and "city"
```scala
from(("id" === id and "name" === name, "city" == city))
``` 
and it composes, (in theory, i haven't test it yet)
you can write for a composite secondary key of "name", "city", "state"
```scala
from((("id" === id  and "name" === name, "city" == city), "state" == state ))
``` 
Let me what you think.  


This is pretty much the "workaround" I used in my codebase, except I used a case class in place of general tuple, which now I think about it isn't very necessary.

chrislewis:
Hello @kailuowang and thanks for taking a look + giving good feedback! I think there's room for improvement and I want to touch on each of the points you raised, but first I want to focus on the the issue around `IndexKey` vs reusing `UniqueKeyCondition`. Just to make sure I've understood correctly, you are suggesting to leave `SecondaryIndex#from` as it was, such that it uses `UniqueKeyCondition` just as `query` does, and then provide additional instances to support values representing composite keys (tuples, for example). If I've misunderstood that, my response here is moot and please correct me.

Assuming I've understood correctly, you are right in that we could simply reuse `UniqueKeyCondition` and provide instances as you suggested, leading to a possible use as you mentioned:

```scala
from((("id" === id  and "name" === name, "city" == city), "state" == state ))
```
With `from` relying on `UniqueKeyCondition`, a good bit of boilerplate in this PR goes away. However, the signature of `from` and `query` would then be the same and, importantly so would their _constraints_. This means that one could then express this invalid query:

```scala
query((("id" === id  and "name" === name, "city" == city), "state" == state ))
```
This is the sole reason why I introduced `IndexKey`: because the constraints of `from` and `query` _must_ be different.

Now, holistically I think the bigger issue is that `UniqueKeyCondition` is serving too many roles (ie, both as a constraint and as a mechanism to bind syntax encodings to the underlying dynamodb API). Such a change would be bigger and would merit more input.

kailuowang:
@chrislewis that's a valid point. I agree that using a separate constraint for `from` can make it more typesafe. 
I am still not sure about the boilerplate for `IndexKey`, that's a lot of code to maintain.  How about we still use the tuples to make it composed. 
Also to keep it more backward compatible with users using custom `UniqueKeyCondition` instance, I suggest that we provide an instance of IndexKey for all `A`s that has a `UniqueKeyCondition`

So the instances can look like this
```scala

implicit def indexKeyConditionTuple[A, B](implicit
      A: IndexKey[A],
      B: IndexKey[B]
  ): IndexKey[(A, B)] =
    new IndexKey[(A, B)] {
      type K = (A.K, B.K)
      final def toDynamoObject(t: (A, B)): DynamoObject =
        A.toDynamoObject(t._1) <> B.toDynamoObject(t._2)
      final def fromDynamoObject(
          key: K,
          dvs: DynamoObject
      ): Option[(A, B)] =
        (A.fromDynamoObject(key._1, dvs), B.fromDynamoObject(key._2, dvs))
          .mapN((_, _))
      final def key(t: (A, B)): K = (A.key(t._1), B.key(t._2))
    }
    

implicit def indexKeyFromUniqkeyCondition[A](implicit
      A: UniqueKeyCondition[A]
  ): IndexKey[A] =
    new IndexKey[A] {
      type K = A.K
      final def toDynamoObject(a: A): DynamoObject =
        A.toDynamoObject(a)

      final def fromDynamoObject(
          key: K,
          dvs: DynamoObject
      ): Option[A] =
       A.fromDynamoObject(key, dvs)

      final def key(a: A): K = A.key(a)
    }
```
Will this work?

chrislewis:
I think you're right about providing instances for any `A` for which there exists `UniqueConditionKey` (in theory that should certainly be true - as you said, it's a superset). I'm away from my laptop for another day or so, so I'll go ahead and mark this a draft and then resume in a day or so. Thanks again for the feedback!

chrislewis:
I see two problems with a tuple-based approach. First is that an implicit provider for a tuple, as in your example:

```scala
implicit def indexKeyConditionTuple[A, B](implicit
      A: IndexKey[A],
      B: IndexKey[B]
  ): IndexKey[(A, B)] = ???
```

would enable invalid instances (and thus invalid keys) to compile. For example, this valid key would compile:


from((("id" === id  and "name" === name, "city" == city), "state" == state ))
```

but so would this (impossible) key:

```scala
from((("id" === id  and "name" === name, "city" == city), "state" == state and "zip" == zip ))
```

Additionally, tuples could be infinitely nested which is another way to generate impossible keys.

Second, even if the tuple-based approach, using nested tuples to represent "and" clauses is difficult to read. I haven't been able to spend enough time on this yet to convince myself of a better way than what's in  this PR, but if nothing else the solution should not permit invalid keys to compile.

kailuowang:
> but so would this (impossible) key:
> 
> ```scala
> from((("id" === id  and "name" === name, "city" == city), "state" == state and "zip" == zip ))
> ```


I am not quite familiar with complex composite key for secondary index, would you expand a bit on why this is impossible?

chrislewis:
Sure thing 👍 

The rules governing the key structure for a table are the same for indexes: there must be at least a partition key and at most a partition key and a sort key (so a "simple" key is just a partition key, while a "complex" key is a partition + sort key).

Secondary indexes can also have keys that differ from the base table: a local index must have the same partition key, but can have a different sort key while a global index can have a different partition and sort key.

Further, the key of an index actually comprises all the components of the base table _and_ those defined on the index. Therefore, the key of an index will contain _at least_ 2 components and _at most_ 4 components. The key in that example I noted has 5 components, which is not possible. Does that make sense?

kailuowang:
Thanks so much! @chrislewis It makes sense. So basically a `from` constraint can have a UniqueKeyCondition for base table key and a `UniqueKeyCondition` for index key right? (Each can be either a `KeyEquals` (partition key only) or `AndEqualsCondition` (partition and sort key)

why not just change the `from` implementation to 

```scala
def from[TableK: UniqueKeyCondition, IndexK: UniqueKeyCondition](tableKey: TableK, indexK: IndexK)
```

we can add an overload to maintain source and binary compatibility but mark it as deprecated 
```scala
def from[TablePartitionK, KeyPartitionK](condition: AndEqualsCondition[TablePartitionK, KeyPartitionK])
```
since this is the only way that old API could have worked for users. 

What do you think?

chrislewis:
No worries, and yes that's right, it could be implemented that way! I hadn't thought of that, and it does have an obvious benefit of needing far less code to implement. However it does come with some downsides:

First is that the change to `from` is not source compatible without overloading it. That's icky, but it would probably work. The implementation of the overloaded `from` would also be a bit more awkward, but not too bad:

```scala
  def from[T: UniqueKeyCondition, I: UniqueKeyCondition](t: UniqueKey[T], i: UniqueKey[I]): SecondaryIndex[V] =
    copy(queryOptions = queryOptions.copy(exclusiveStartKey = Some(t.toDynamoObject <> i.toDynamoObject)))
```
Second is that _extracting_ the key for an index from a query response is even more awkward that it already is to extract a key for a table. Consider this test code, which is expanded from an already existing example to demonstrate index paging and access to the query result on an index:

```scala
it("Scans the table and returns the raw DynamoDB result") {

    LocalDynamoDB.withRandomTableWithSecondaryIndex(client)("mode" -> S, "line" -> S)("mode" -> S, "colour" -> S) {
      (t, i) =>
        val table = Table[Transport](t)
        val index = table.index(i)
        val ops = for {
          _ <- table.putAll(
            Set(
              Transport("Underground", "Circle", "Y"),
              Transport("Underground", "Metropolitan", "M"),
              Transport("Underground", "Central", "R")
            )
          )
          res <- index.limit(1).scan0
          tableKey =
            UniqueKeyCondition[AndEqualsCondition[KeyEquals[String], KeyEquals[String]], (AttributeName, AttributeName)]
          tk = tableKey.fromDynamoObject(("mode", "line"), DynamoObject(res.lastEvaluatedKey))
          ik = tableKey.fromDynamoObject(("mode", "colour"), DynamoObject(res.lastEvaluatedKey))
          keys = tk.flatMap(t => ik.map(t -> _))
          ts <- keys.fold(List.empty[Either[DynamoReadError, Transport]].pure[ScanamoOps]) {
            case (t, i) => index.from(t, i).scan()
          }
        } yield ts
        scanamo.exec(ops) should be(
          List(Right(Transport("Underground", "Central", "R")), Right(Transport("Underground", "Circle", "Y")))
        )
    }
  }
```

The summoning of the `UniqueKeyCondition` assigned to `tableKey` is pretty unwieldy, and that's what's already required just to extract a paging key from a raw response. Because there are now two distinct `UniqueKeyCondition` instances required to page this index, we must extract two different objects from the result (here named `tk` and `ik`). The need to extract two already seems more confusing, but to make it worse, imagine if the index key only had one non-overlapping component (that is, imagine the table key fit `AndEqualsCondition[KeyEquals[String], KeyEquals[String]]` but the index key fit `KeyEquals[String]`. In that case one would have to summon an additional `UniqueKeyCondition` of the correct shape, which makes it even more unwieldy, confusing, and easy to get wrong.

Paging in scanamo already feels like an awkward second class citizen, and while I appreciate that this approach would at least fully enable paging over indexes without enabling invalid queries, I think it would do so at a significant cost in user experience. I think most would agree that the difference in the code size of this approach (almost nil) compared to this pull request is significant, and that is a valid criticism. However I don't think the choice of implementation should come at the cost of the user experience, and I think this approach would clearly do exactly that.

kailuowang:
@chrislewis first of all, thank you so much for bearing with me. We are approaching 1.0 release, so I am a bit extra cautious in API and code decisions. 
 I 100% agree that we should make it easier for users to extract key back from the response, And I also agree that the current API makes "paging in scanamo already feels like an awkward second class citizen" and we definitely want to keep the `scan0` and `query0` you added, but I'd suggest naming them `scanRaw` `queryRaw` 


Then we need to come up with some better API to extract lastEvaluatedKey 
The problem using `UniqueKeyCondition[T]` here is that the type function is from value type (e.g. KeyEquals) to key type  (e.g. `AttributeName`). For extracting response we need to opposite, i.e. from Key type e.g. `AttributeName` and `V` to `KeyEquals`. 
We need some type computation helper. The following is my experiementation:

```scala
trait UniqueKeyExtractor[A] {
  type T
  type K
  def uc: UniqueKeyCondition.Aux[T, K]
  def apply(a: A): K
}
case class Attribute[V](name: String)

import org.scanamo.query.AttributeName
object  UniqueKeyExtractor {
  implicit def attributeInstance[V]
   (implicit summoned: UniqueKeyCondition.Aux[KeyEquals[V], AttributeName]): UniqueKeyExtractor[Attribute[V]] =
     new UniqueKeyExtractor[Attribute[V]] {
    type T = KeyEquals[V]
    type K = AttributeName
    val uc = summoned
    def apply(a: Attribute[V]): AttributeName = a.name

  }

  implicit def andEqualInstance[V1, V2]
   (implicit summoned: UniqueKeyCondition.Aux[AndEqualsCondition[KeyEquals[V1], KeyEquals[V2]], (AttributeName, AttributeName)]): UniqueKeyExtractor[(Attribute[V1], Attribute[V2])] =
     new UniqueKeyExtractor[(Attribute[V1], Attribute[V2])] {
    type T = AndEqualsCondition[KeyEquals[V1], KeyEquals[V2]]
    type K = (AttributeName, AttributeName)
    val uc = summoned
    def apply(a: (Attribute[V1], Attribute[V2])): (AttributeName, AttributeName) = (a._1.name, a._2.name)
   }
}
```
Then we can decare the following method somewhere
```scala
  import cats.implicits._
  def extractIndexKey[TK, IK](tk: TK, ik: IK, dynamoObject: DynamoObject)
                             (implicit TK: UniqueKeyExtractor[TK], IK: UniqueKeyExtractor[IK]): Option[(TK.T, IK.T)] = {
    (TK.uc.fromDynamoObject(TK(tk), dynamoObject), IK.uc.fromDynamoObject(IK(ik), dynamoObject)).tupled  
  }

```
Then we can call it like 
```scala
  extractIndexKey(Attribute[String]("A"), Attribute[String]("B"), aDynamoObject)
  
  extractIndexKey((Attribute[String]("A"), Attribute[String]("B")), Attribute[String]("C"), dynamoObject)

```

if we really don't like using `tuple` to represent partKey and sortKey, then we can create a spellout data struct out
to represent a key

```scala
case class ComplexKey[PV, SV](partKey: Attribute[V], sortKey: Attribute[SV])
```

And provide an `UniqueKeyExtractor` insetance for it like the one provided for tuple
then usage can read as 

```scala
  extractIndexKey(ComplexKey(Attribute[String]("A"), Attribute[String]("B")), Attribute[String]("C"), dynamoObject)
```
I am sure there are more ways to improve this. But I really think we should spell out the dynamo key structure with explicit types, rather than a var arg like api that takes 2, 3, or 4 arguments.

chrislewis:
@kailuowang no worries, I fully understand and support the lean towards stability in the API of an established library. At the same time, if scanamo is going to be an option for my team (and I would bet many others), it needs first-class support for paging (including indexes) on top of the safety claims it purports to offer. I believe it can get there, but it's not clear that such a change will be tiny. I'd be happy to help with that effort in terms of code, ideas, opinions, or even offer feedback on the ideas from others.

As for how to move forward, you mention keeping `scan0` + `query0`. If this were added it would at least shrink the surface area of hacks people need to layer over scanamo to be able to paginate indexes. I don't love the idea of maintaining our own (potentially dangerous) instances of `UniqueKeyCondition` in order to use `SecondaryIndex#from` as-is, but at least we could move forward.

I'd be glad to open another PR with just `scan0` + `query0`. I don't really care about the name suffix `0` vs `Raw`, but I do care about consistency in an API (which is why I named them `scan0` and `query0`, as they are so named on `Table`). It would be awkward to have methods on `SecondaryIndex` serving essentially the same purpose as ones on `Table`, but with the former using a different naming convention. Anyway, let me know if another PR for those is acceptable and we can chat about that there.

Final half-baked thoughts about how pagination might look in the (hopefully near) future -->

For operations that cannot be guaranteed to return the complete result set targeted by the operation (ie query, scan), it seems odd to model the results of such operations merely as collections of result values. I would argue that pagination is a fundamental feature of dynamodb given that data will automatically (and forcibly) be paged if the result set exceeds 1MB. Because of that it seems very odd to even need access to the raw response just for the sake of accessing the page key; I would argue that the possibility of receiving a paginated response should be _baked into_ the result type. That has some API and modeling ramifications that need digging into (again, happy to help), but it seems only natural to have such first class support. Last of all I want to be clear that I am not implying that access to the raw response should be removed given it may be useful for various as-hoc/use-case-specific purposes, but I am asserting that pagination should not require such low-level access.

kailuowang:
Yes, separating out a PR for scan0 and query0 sounds a good idea. I still prefer naming them scanRaw and queryRaw and add the corresponding aliases to `Table` to keep it consistent. 

>  I am asserting that pagination should not require such low-level access.
I agree that we should design a new method in the API for native pagination support.  Arguably the SecondaryIndex API right now is pretty much broken. I would prefer we deprecate the old method but keep it in 1.0, but what to do with the old method can be left for another discussion.

PR #1346, opened 2022-01-25
Add SecondaryIndex methods exposing raw response

chrislewis:
`Table` has provides `scan0` and `query0` to access the raw response, for which a common use-case is to access the last evaluated (ie paging) key. `SecondaryIndex` lacks these methods, making it impossible to page indexes. Per #1324, this slimmed-down PR adds those methods so that paged queries are at least possible on indexes. However the larger issue around paging indexes (ie, the impossibility to safely form a complete index paging key in common use-cases) still stands. This PR is at least a workaround until the larger issue can be addressed.

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
+++ b/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
@@ -196,7 +196,10 @@ object ScanamoFree {
   ): ScanamoOpsT[M, List[Either[DynamoReadError, T]]] =
     ScanResponseStream.streamTo[M, T](ScanamoScanRequest(tableName, None, ScanamoQueryOptions.default), pageSize)
 
-  def scan0[T: DynamoFormat](tableName: String): ScanamoOps[ScanResponse] =
+  @deprecated("use `scanRaw`", "1.0")
+  def scan0[T: DynamoFormat](tableName: String): ScanamoOps[ScanResponse] = scanRaw[T](tableName)
+
+  def scanRaw[T: DynamoFormat](tableName: String): ScanamoOps[ScanResponse] =
     ScanamoOps.scan(ScanamoScanRequest(tableName, None, ScanamoQueryOptions.default))
 
   def query[T: DynamoFormat](tableName: String)(query: Query[_]): ScanamoOps[List[Either[DynamoReadError, T]]] =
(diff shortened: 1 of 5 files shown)


kailuowang (approved):
Thanks so very much!

chrislewis:
Yw! Out of curiosity, what is the protocol for how PRs are merged and how that feeds into a release? I only recently found [this post](https://github.com/scanamo/scanamo/releases/tag/v1.0-M13) by @regiskuckaertz, so it seems he isn't really involved while at the same time is the only person listed in `CODEOWNERS`.

Is there somewhere that describes the release cycle, or even a place where it can be discussed? The gitter channels (there are [two](https://gitter.im/guardian/scanamo) of [them](https://gitter.im/scanamo/scanamo)) both seem dead or at least sparsely active.

chrislewis:
Sounds good ty!

kailuowang:
> Sounds good ty!

Let's give it another day or 2 to see if any other maintainers have have feedback

scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
```diff
@@ -196,7 +196,10 @@ object ScanamoFree {
   ): ScanamoOpsT[M, List[Either[DynamoReadError, T]]] =
     ScanResponseStream.streamTo[M, T](ScanamoScanRequest(tableName, None, ScanamoQueryOptions.default), pageSize)
 
-  def scan0[T: DynamoFormat](tableName: String): ScanamoOps[ScanResponse] =
+  @deprecated("use `scanRaw`", "1.0")
+  def scan0[T: DynamoFormat](tableName: String): ScanamoOps[ScanResponse] = scanRaw[T](tableName)
+
+  def scanRaw[T: DynamoFormat](tableName: String): ScanamoOps[ScanResponse] =
```
regiskuckaertz, line 202: this is way better 😅

regiskuckaertz (approved):
😍 this will make auto-complete way more user-friendly.

kailuowang:
@chrislewis we ran out of Travis credit, a new release requires migrating to GitHub action now
https://github.com/scanamo/scanamo/issues/1356
As soon as we get the secrets set up, I will release the next version

chrislewis:
🙌 👀

rtyley:
As a stop-gap, I just got CI running again with Travis for now- this shows that unfortunately, there was a linting problem with this PR:

https://app.travis-ci.com/github/scanamo/scanamo/jobs/558769630#L286-L290

```
[warn] /home/travis/build/scanamo/scanamo/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala isn't formatted properly!
[warn] /home/travis/build/scanamo/scanamo/scanamo/src/main/scala/org/scanamo/SecondaryIndex.scala isn't formatted properly!
```

...so the CI build is now running, but failing! Would you be able to fix-up the formatting? I'm technically away from work until Thursday, but will try to look into helping out with the CI migration to GitHub Actions.

rtyley:
> fix-up the formatting (run `scalafmt` and then `scalafmtCheck` to check if the files pass)\

Done with 03e42fae3fc2a51d53bac4f8bc2af8b601673e97, and released as ([`v1.0.0-M19`](https://github.com/scanamo/scanamo/releases/tag/v1.0.0-M19), [here](https://repo1.maven.org/maven2/org/scanamo/scanamo_2.13/) on Maven Central)!

chrislewis:
Nice - thanks for all the patch ups! Incidentally we actually do use scalafmt internally, but we wire it in to the build (as in build tool, not CI) so compilation forces formatting (precisely to avoid issues like this), and then we more or less forget about it. I realize that can feel intrusive to some, but I've found that it helps move teams beyond the time-consuming debates about how code should be formatted.

Thanks again all!

PR #1570, opened 2022-08-28
add support for zio 2

DannyMor:
Changes made: 
1. update zio and zio-streams version to 2.0.1
2. update zio-interop-cats to 3.3.0
3. update scanamoZio with new ZIO Streams api call
4. update ZioInterpreter with new ZIO effects api
5. update alpakka tests to gracefully ignore non unit return value
6. update ZIO effects execution for ZIO tests

How it was tested:
ran all tests against local dynamodb 

ran contribution guidelines commands as well

### DIFF
--- a/alpakka/src/test/scala/org/scanamo/ScanamoAlpakkaSpec.scala
+++ b/alpakka/src/test/scala/org/scanamo/ScanamoAlpakkaSpec.scala
@@ -43,11 +43,11 @@ class ScanamoAlpakkaSpec extends AnyFunSpecLike with BeforeAndAfterAll with Matc
 
       scanamo
         .exec(result)
-        .runForeach(
-          _ should equal(
+        .runForeach { x =>
+          val _ = x should equal(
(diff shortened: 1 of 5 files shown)


DannyMor:
@regiskuckaertz Hi, is there a chance to add support for zio2 in scanamo? if so, it would be amazing!
I did get a workaround by using the DynamoDbAsyncClient but just wanted to know if it's planned for the near future.
BTW, Scanamo is amazing, thank you for that!!!

paul-danilin-moia:
Hi, we are also interested in the zio 2 support. @kailuowang could you please take a look?
Btw, zio 2 support could live in a separate module if we wanted to support both zio 1 and zio 2 users. I don't mind simply updating the existing module to zio 2 though.

joroKr21:
Another take: https://github.com/scanamo/scanamo/pull/1514
I think at this point it's better to just upgrade because ZIO 1 libraries are not maintained in general.

PR #1610, opened 2022-12-01
fix: share the source file across scala 2 and 3

i10416:
This PR removes Scala 3 patch file for kind-projector syntax.
Scala 3 uses `_` for type lambda placeholders and there is scalac option to use `_` as type lambda placeholders in Scala 2.

For details, see https://docs.scala-lang.org/scala3/guides/migration/plugin-kind-projector.html


kailuowang (approved):

kailuowang:
Thanks very much!

PR #1622, opened 2023-01-28
Fix NPE in DynamoResultStream

eugene-cheverda:
The NPE occurs when scannedCount underlying field is null. Simple map on Option doesn't help as apparently there's implicit cast from Java Integer  to Scala Int happening and it would still fail with NPE, therefore chain of flatMaps.

It would be great if merge of this PR would result in new build published to Maven. Thanks in advance!


regiskuckaertz (approved):
❤️ thanks

PR #1629, opened 2023-02-20
Add support for multi-action write transactions

chrischivers:
This PR provides support for multi-action write requests (as detailed in the Dynamo spec [here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html))

Currently the Scanamo API allows only write requests of the same type (Put, Update, Delete) to be grouped together in a single transaction (i.e. `transactPutAll`, `transactUpdateAll`, `transactDeleteAll`).

This change introduces `transactionWrite` which accepts multiple TransactionWriteActions. These actions could be a mix of `Put`, `Update`, `Delete` or `ConditionCheck` types and can span multiple tables.

### DIFF
--- a/cats/src/main/scala/org/scanamo/ops/CatsInterpreter.scala
+++ b/cats/src/main/scala/org/scanamo/ops/CatsInterpreter.scala
@@ -92,6 +92,7 @@ class CatsInterpreter[F[_]](client: DynamoDbAsyncClient)(implicit F: Async[F]) e
               a => F.delay(Right(a))
             )
           )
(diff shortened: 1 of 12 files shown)


kailuowang (approved):
This looks great to me. It would be even nicer if we could add some docs but that shouldn't be holding this PR being merged.

kailuowang:
Summoning @rtyley to see if you have some time to take a quick look.

regiskuckaertz (approved):

PR #1643, opened 2023-05-03
Update ci to include doc, downgrade scala due to dotty bug

kailuowang:
Released failed due to a [`Scala 3.2.1+` bug](https://github.com/lampepfl/dotty/issues/15288) causing `doc` to fail on Scala 3.2.2
This bug is fixed and will likely be released 3.3.1


rtyley (approved):
Looks good- nice one!

PR #1647, opened 2023-05-22
cats interpreter should return a finaliser. v3.5.0 changed the  Cancelation Semantics

inanme:
related to https://github.com/typelevel/cats-effect/releases/tag/v3.5.0


kailuowang (approved):

kailuowang:
Thank you!

Issue #819, opened 2020-08-16
Avoid retries of retries


RoyCarbyne:
Alpakka using AWS SDK now, which use AWS best practices include retries internally 
RetryPolicy perform retries on Alpakka

We have reties of retries now
I suggest to delete RetryPolicy

regiskuckaertz:
Great point! Thanks for the heads up, I will remove.

PR #1790, opened 2024-07-20
Extract common `exec` & `execT` methods into new `ScanamoClient`

rtyley:
The 5 Scanamo client objects (`Scanamo`, `ScanamoAsync` `CatsScanamo`, `ZioScanamo`, `PekkoScanamo`) all have common `exec` & `execT` methods with almost completely identical implementations. This change introduces a new common `ScanamoClient` super-class for all 5 client objects that implements those methods.

Result: less code! ✨ 

image

### Do we need the subtype instance classes?!

I was tempted to remove the 5 subtype instance classes altogether, as there's not much need for them from the Scanamo implementation side - we could get by with just the 5 singleton objects providing `apply()` methods to construct `ScanamoClient` instances - but there are some wrinkles with that:

* Existing code in consumers of the Scanamo library will typically hand around instance references with the type of the subtype (ie `scanamo: ScanamoAsync`) - they've had no alternative before. We can't remove the subtype instance class without forcing a lot of users to switch to using `scanamo: ScanamoClient[Future]` (which is also less readable!). Swapping in some type aliases, eg `type ScanamoAsync = ScanamoClient[Future]` might help here, but I won't try that here.
* `PekkoScanamo` provides an additional `execFuture` method, which is Pekko-specific - no idea if it's used, but I guess we shouldn't delete it.
* Although `ScanamoClient[Id]` should be able to cleanly implement `Scanamo` without alteration (because `type Id[A] = A`), it turns out that this would mean that _some_ consuming code would have to be altered for the Scala 2 compiler (the Scala 3.3.3 compiler did not have this problem).

...so the 5 subtype instance class definitions continue to exist for now.

### See also

* https://github.com/scanamo/scanamo/pull/1786 for a subsequent refactor that also reduces code.

### DIFF
--- a/cats/src/main/scala/org/scanamo/ScanamoCats.scala
+++ b/cats/src/main/scala/org/scanamo/ScanamoCats.scala
@@ -17,19 +17,12 @@
 package org.scanamo
 
 import cats.effect.Async
-import cats.{ ~>, Monad }
+import cats.~>
 import fs2.Stream
-import org.scanamo.ops.{ CatsInterpreter, ScanamoOps, ScanamoOpsT }
+import org.scanamo.ops.CatsInterpreter
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
 
-class ScanamoCats[F[_]: Async](client: DynamoDbAsyncClient) {
-  final private val interpreter = new CatsInterpreter(client)
-
-  final def exec[A](op: ScanamoOps[A]): F[A] = op.foldMap(interpreter)
-
-  final def execT[M[_]: Monad, A](hoist: F ~> M)(op: ScanamoOpsT[M, A]): M[A] =
-    op.foldMap(interpreter andThen hoist)
-}
+class ScanamoCats[F[_]: Async] private (client: DynamoDbAsyncClient) extends ScanamoClient(new CatsInterpreter(client))
 
 object ScanamoCats {
   def apply[F[_]: Async](client: DynamoDbAsyncClient): ScanamoCats[F] = new ScanamoCats(client)
--- a/pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
+++ b/pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
@@ -16,74 +16,45 @@
 
 package org.scanamo
 
-import cats.{ ~>, Monad }
+import cats.Monad
 import org.apache.pekko.NotUsed
 import org.apache.pekko.actor.ClassicActorSystemProvider
 import org.apache.pekko.stream.scaladsl.{ Sink, Source }
-import org.scanamo.ops.PekkoInterpreter.Pekko
-import org.scanamo.ops.{ PekkoInterpreter, ScanamoOps, ScanamoOpsT }
+import org.scanamo.PekkoInstances.*
+import org.scanamo.ops.{ PekkoInterpreter, ScanamoOps }
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
-import software.amazon.awssdk.services.dynamodb.model.*
 
 import scala.concurrent.Future
 
-/** Provides the same interface as [[org.scanamo.Scanamo]], except that it requires an
-  * [[https://github.com/apache/incubator-pekko-connectors Pekko connectors]] client, a
-  * [[org.scanamo.ops.retrypolicy.RetryPolicy]] and a predicate for which [[scala.Throwable]]s should be retried.
-  * `retryPolicy` defaults to [[org.scanamo.ops.retrypolicy.RetryPolicy.max]] with maximum 3 retries if not explicitly
-  * provided. `isRetryable` defaults to retry the most common retryable Dynamo exceptions. Moreover, the interface
-  * returns either a [[scala.concurrent.Future]] or [[org.apache.pekko.stream.scaladsl.Source]] based on the kind of
-  * execution used.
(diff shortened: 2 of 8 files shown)


pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
```diff
@@ -16,74 +16,43 @@
...
-    throwable match {
-      case _: InternalServerErrorException | _: ItemCollectionSizeLimitExceededException | _: LimitExceededException |
-          _: ProvisionedThroughputExceededException | _: RequestLimitExceededException =>
-        true
-      case e: DynamoDbException
-          if e.awsErrorDetails.errorCode.contains("ThrottlingException") |
-            e.awsErrorDetails.errorCode.contains("InternalFailure") =>
-        true
-      case _ => false
-    }
```
rtyley, line 73: This `defaultRetryableCheck()` method, originally introduced with https://github.com/scanamo/scanamo/pull/584 in January 2020, has been obsolete since https://github.com/scanamo/scanamo/pull/825 removed Scanamo's superfluous retry logic in August 2020, as per https://github.com/scanamo/scanamo/issues/819.

scanamo/src/main/scala/org/scanamo/ScanamoSync.scala
```diff
@@ -16,25 +16,16 @@
...
-  /** Execute the operations built with [[org.scanamo.Table]]
-    */
-  def exec[A](op: ScanamoOps[A]): A = op.foldMap(interpreter)
-
-  /** Execute the operations built with [[org.scanamo.Table]] with effects in the monad `M` threaded in.
-    */
-  def execT[M[_]: Monad, A](hoist: Id ~> M)(op: ScanamoOpsT[M, A]): M[A] =
-    op.foldMap(interpreter andThen hoist)
+final class Scanamo private (client: DynamoDbClient) extends ScanamoClient(new ScanamoSyncInterpreter(client)) {
+  override def exec[A](op: ScanamoOps[A]): A = super.exec(op) // overridden to explicitly return `A` rather than `Id[A]`
```
rtyley, line 28: This override is necessary to ensure the _"Filters a table for attributes containing a specific substring"_ test (and presumably, other consumer code out in the wild) compiles under Scala 2.

pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
```diff
@@ -16,74 +16,43 @@
...
 
-/** Provides the same interface as [[org.scanamo.Scanamo]], except that it requires an
-  * [[https://github.com/apache/incubator-pekko-connectors Pekko connectors]] client, a
-  * [[org.scanamo.ops.retrypolicy.RetryPolicy]] and a predicate for which [[scala.Throwable]]s should be retried.
-  * `retryPolicy` defaults to [[org.scanamo.ops.retrypolicy.RetryPolicy.max]] with maximum 3 retries if not explicitly
-  * provided. `isRetryable` defaults to retry the most common retryable Dynamo exceptions. Moreover, the interface
-  * returns either a [[scala.concurrent.Future]] or [[org.apache.pekko.stream.scaladsl.Source]] based on the kind of
-  * execution used.
+/** The interface returns either a [[scala.concurrent.Future]] or [[org.apache.pekko.stream.scaladsl.Source]] based on
+  * the kind of execution used.
```
rtyley, line 30: The comments about retry logic are obsolete since https://github.com/scanamo/scanamo/pull/825 removed Scanamo's superfluous retry logic in August 2020, as per https://github.com/scanamo/scanamo/issues/819.

pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
```diff
@@ -16,74 +16,43 @@
...
-
-  final def execT[M[_]: Monad, A](hoist: Pekko ~> M)(op: ScanamoOpsT[M, A]): M[A] =
-    op.foldMap(interpreter.andThen(hoist))
-
-  def execFuture[A](op: ScanamoOps[A]): Future[A] =
-    run(op).runWith(Sink.head[A])
-
-  private def run[A](op: ScanamoOps[A]): Pekko[A] =
-    op.foldMap(interpreter)
+  def execFuture[A](op: ScanamoOps[A]): Future[A] = exec(op).runWith(Sink.head[A])
```
rtyley, line 38: `PekkoScanamo` has provided this additional `execFuture` method, which is Pekko-specific - no idea if it's used in the wild, but I guess we shouldn't delete it.

cats/src/main/scala/org/scanamo/ScanamoCats.scala
```diff
@@ -17,19 +17,12 @@
...
 
-class ScanamoCats[F[_]: Async](client: DynamoDbAsyncClient) {
-  final private val interpreter = new CatsInterpreter(client)
-
-  final def exec[A](op: ScanamoOps[A]): F[A] = op.foldMap(interpreter)
-
-  final def execT[M[_]: Monad, A](hoist: F ~> M)(op: ScanamoOpsT[M, A]): M[A] =
-    op.foldMap(interpreter andThen hoist)
-}
+class ScanamoCats[F[_]: Async] private (client: DynamoDbAsyncClient) extends ScanamoClient(new CatsInterpreter(client))
```
rtyley, line 26: Note that, in line with the other Scanamo clients, the constructor is now made `private`, and you need to use the companion object `apply()` method to create the client. So you can no longer use:

```scala
new ScanamoCats(dynamoDbAsyncClient)
```

...but instead just:

```scala
ScanamoCats(dynamoDbAsyncClient)
```

PR #1796, opened 2024-08-13
Simplify processing of `RequestCondition` attribute values - make them non-optional

rtyley:
* **`RequestCondition` dynamo values become non-optional**: instead they can just be `DynamoObject.empty`
* **New class `AttributeNamesAndValues`**: we often combine the different sources of attribute names/values in a request (eg, 'update' and 'condition') and there's no point in repeatedly writing code where names are combined, and then _values_ are combined - we always want to aggregate both at the same time, now supported with `AttributeNamesAndValues`
* **New class `UpdateAndCondition`**: combines `UpdateExpression` & `Option[Condition]` - models the payload of both normal `UpdateItem` & transactional `Update` requests.
* Using the [lambda syntax for Single Abstract Method (SAM) types](https://www.scala-lang.org/news/2.12.0/#lambda-syntax-for-sam-types) in the implementation of many typeclass instances of `ConditionExpression` and `QueryableKeyCondition` (eg `QueryableKeyCondition[KeyEquals[V]]`) makes the implementations much shorter.

The 'values' on `RequestCondition` have been optional since `RequestCondition` was introduced with https://github.com/scanamo/scanamo/pull/31 in May 2016 - back then they were `attributeValues: Option[Map[String, AttributeValue]]`), later becoming `dynamoValues: Option[DynamoObject]` with https://github.com/scanamo/scanamo/pull/400 in May 2019.

The values are probably 'optional' because there are certain DynamoDB condition expression functions (like `attribute_exists()`) that _don't_ take a value, so it was _reasonable_ for implementations like `ConditionExpression[AttributeExists]` to create instances of `RequestCondition` where `dynamoValues` was `None`:

https://github.com/scanamo/scanamo/blob/93ab1e7e8292c79283b556e00389017beb8559c0/scanamo/src/main/scala/org/scanamo/query/ConditionExpression.scala#L180-L184

However, if we make `dynamoValues` non-optional, we can just supply `DynamoObject.empty` rather than `None`- and this allows us to simplify so much logic around combining the attribute values that come from different sources (eg as conditions are combined). Note that Cats [`Monoid`](https://typelevel.org/cats/typeclasses/monoid.html) typeclass can provide useful functionality like [`.orEmpty`](https://typelevel.org/cats/api/cats/syntax/OptionOps.html#orEmpty(implicitA:cats.Monoid[A]):A) that helps with this.

### Classes with field changes

* **`ScanamoUpdateRequest`**: `updateExpression`, `attributeNames`, `dynamoValues`, `addEmptyList`, `condition` are all replaced by `UpdateAndCondition`
* **`TransactUpdateItem`**: `updateExpression`, `condition` are replaced by `UpdateAndCondition`
* **`RequestCondition`**: `attributeNames`, `dynamoValues` are replaced by `AttributeNamesAndValues`

### Affected code using Scanamo

Scanamo doesn't have a clear divide between internal code and external API, but `org.scanamo.request.RequestCondition` is not _explicitly_ part of the public API - it's not mentioned in the documentation at [www.scanamo.org/conditional-operations](https://www.scanamo.org/conditional-operations) for instance. Changing the type of `RequestCondition.dynamoValues` doesn't feel like it should break any user-code - it certainly didn't require any changes to Scanamo's tests, for instance.

However, doing a [search on GitHub](https://github.com/search?q=org.scanamo.request.RequestCondition+NOT+repo%3Ascanamo%2Fscanamo+NOT+is%3Afork&type=code) finds recent code in [github.com/nationalarchives](https://github.com/nationalarchives) that will break on this change - both call `requestCondition.dynamoValues.flatMap()`:

* https://github.com/nationalarchives/dr2-ingest/blob/9954492b605973b80f707be2fc5bb88981cc3889/scala/lambdas/ingest-files-change-handler/src/test/scala/uk/gov/nationalarchives/ingestfileschangehandler/Utils.scala#L138
* https://github.com/nationalarchives/da-aws-clients/blob/f46ad91eaca46f9ba293763f3da75e3ddd070c1f/dynamodb/src/main/scala/uk/gov/nationalarchives/DADynamoDBClient.scala#L248

### Support for legacy code that calls removed methods

For all 3 classes where fields have been removed with a better implementation, read-only _deprecated_ versions of those fields are now provided - so, for instance, you _can_ call `requestCondition.dynamoValues.flatMap()`, but `dynamoValues` will be deprecated.

Additionally, for each of `TransactUpdateItem` & `RequestCondition`, an [auxiliary constructor](https://docs.scala-lang.org/scala3/book/domain-modeling-tools.html#auxiliary-constructors) has been added which maintains the original constructor signature as a deprecated constructor. This wasn't done for `ScanamoUpdateRequest` as re-creating the `UpdateExpression` from the legacy arguments seemed Too Hard.

These deprecated methods will be left in the code for at least a year after this change is released - but they will deleted eventually!

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/request/ScanamoRequest.scala
+++ b/scanamo/src/main/scala/org/scanamo/request/ScanamoRequest.scala
@@ -16,8 +16,9 @@
 
 package org.scanamo.request
 
+import cats.implicits.*
 import org.scanamo.query.{ Condition, Query }
-import org.scanamo.update.UpdateExpression
+import org.scanamo.update.{ UpdateAndCondition, UpdateExpression }
 import org.scanamo.{ DeleteReturn, DynamoObject, DynamoValue, PutReturn }
 
 case class ScanamoPutRequest(
@@ -37,12 +38,19 @@ case class ScanamoDeleteRequest(
 case class ScanamoUpdateRequest(
   tableName: String,
   key: DynamoObject,
-  updateExpression: String,
-  attributeNames: Map[String, String],
-  dynamoValues: DynamoObject,
-  addEmptyList: Boolean,
-  condition: Option[RequestCondition]
-)
+  updateAndCondition: UpdateAndCondition
+) {
+  @deprecated("See https://github.com/scanamo/scanamo/pull/1796", "3.0.0")
+  def updateExpression: String = updateAndCondition.update.expression
+  @deprecated("See https://github.com/scanamo/scanamo/pull/1796", "3.0.0")
+  def attributeNames: Map[String, String] = updateAndCondition.update.attributes.names
+  @deprecated("See https://github.com/scanamo/scanamo/pull/1796", "3.0.0")
+  def dynamoValues: DynamoObject = updateAndCondition.update.attributes.values
+  @deprecated("See https://github.com/scanamo/scanamo/pull/1796", "3.0.0")
+  def addEmptyList: Boolean = updateAndCondition.update.addEmptyList
+  @deprecated("See https://github.com/scanamo/scanamo/pull/1796", "3.0.0")
+  def condition: Option[RequestCondition] = updateAndCondition.condition
+}
 
 case class ScanamoScanRequest(
   tableName: String,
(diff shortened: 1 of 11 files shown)


scanamo/src/main/scala/org/scanamo/request/ScanamoRequest.scala
```diff
@@ -71,7 +71,7 @@ object ScanamoQueryOptions {
 case class RequestCondition(
   expression: String,
   attributeNames: Map[String, String],
-  dynamoValues: Option[DynamoObject]
+  dynamoValues: DynamoObject
```
rtyley, line 74: This is the change to the model that drives all the simplifications we see in the rest of the PR.

PR #1799, opened 2024-08-19
Extend AWS SDK builders to ease `Option`-setting

rtyley:
Scanamo's own Scala-based models of DynamoDB requests use [Scala's idiomatic approach to optional values](https://alvinalexander.com/scala/using-scala-option-some-none-idiom-function-java-null/) - use `Option` - but this repeatedly gives us a fiddly job when coping with the [fluent nature of the AWS SDK for Java](https://aws.amazon.com/blogs/developer/fluent-client-builders/), which chains methods on builders. We want to pretend that the builders aren't mutable (because if they were written in Scala, they wouldn't be mutable), and using the default methods on Option, the most concise way to do that for a builder is using `optionValue.foldLeft(builder)(builder.setter)` - which is _reasonably_ concise, but you'll notice that we're repeating `builder` twice, and it's not possible to chain together invocations of this - although it _produces_ an instance of builder, it doesn't _start_ from an instance of builder, so: no chaining.

This change introduces a `setOpt` extension method on all AWS SDK `Buildable`s:

https://github.com/scanamo/scanamo/blob/79370f2d09ad941d5fe25075ea8437498cb06db2/scanamo/src/main/scala/org/scanamo/internal/aws/sdkv2/package.scala#L6-L8

This provides a more concise optional-invocation of the builder's setter methods, while supporting the same fluent chaining that native builder setter methods use. `f: B => (V => B)` is used because [partial application](https://docs.scala-lang.org/tour/multiple-parameter-lists.html#partial-application) of the builder setter method in many cases allows us to use a single underscore rather than two - ie `.setOpt(req.index)(_.indexName)` rather than `.setOpt(req.index)(_ indexName _)` which we'd use if we had `f: (B, V) => B`.

Because the builder type `B` is a type-parameter, it's also more flexible than the `queryRefinement()` method that was duplicated in `scan(req: ScanamoScanRequest)` & `query(req: ScanamoQueryRequest)` until now:

https://github.com/scanamo/scanamo/blob/dcc4b4bf35904a0a4b6324525ef3ffd834db27e1/scanamo/src/main/scala/org/scanamo/ops/package.scala#L33-L37

...those can now be replaced with the single `setOpt()` implementation.

The result is code that's more readable, with a good reduction in line-count! ✨ 

image

# Inspecting the diff

This diff is best inspected in split configuration, with whitespace changes ignored: https://github.com/scanamo/scanamo/pull/1799/files?diff=split&w=1

I've gone through and visually inspected to check that all setting-of-fields has been copied over - of course, we hope that the tests would have caught it if any field setting _had_ been lost.

### DIFF
--- a/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
+++ b/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
@@ -16,6 +16,7 @@
 
 package org.scanamo
 
+import cats.implicits.*
 import cats.{ Monad, MonoidK }
 import org.scanamo.DynamoResultStream.{ QueryResponseStream, ScanResponseStream }
 import org.scanamo.ops.ScanamoOps.Results.*
@@ -70,7 +71,7 @@ object ScanamoFree {
           batch,
           item =>
             WriteRequest.builder
-              .putRequest(PutRequest.builder.item(f.write(item).asObject.getOrElse(DynamoObject.empty).toJavaMap).build)
+              .putRequest(PutRequest.builder.item(f.write(item).asObject.orEmpty.toJavaMap).build)
               .build
         )
       }
--- a/scanamo/src/main/scala/org/scanamo/internal/aws/sdkv2/package.scala
+++ b/scanamo/src/main/scala/org/scanamo/internal/aws/sdkv2/package.scala
@@ -1,0 +1,11 @@
+package org.scanamo.internal.aws
+
+import software.amazon.awssdk.utils.builder.Buildable
+
+package object sdkv2 {
+  implicit class RichBuilder[B <: Buildable](builder: B) {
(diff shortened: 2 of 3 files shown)


scanamo/src/main/scala/org/scanamo/ScanamoFree.scala
```diff
@@ -70,7 +71,7 @@ object ScanamoFree {
           batch,
           item =>
             WriteRequest.builder
-              .putRequest(PutRequest.builder.item(f.write(item).asObject.getOrElse(DynamoObject.empty).toJavaMap).build)
+              .putRequest(PutRequest.builder.item(f.write(item).asObject.orEmpty.toJavaMap).build)
```
rtyley, line 74: Sneaking this unrelated change into this PR - should have done it when we introduced `Monoid[DynamoObject]` in #1796.

PR #1801, opened 2024-08-23
support enabling time to live feature

prakhar-flow:


rtyley (commented):
Hi @prakhar-flow, thank you for your contribution! Unfortunately, I'm uncertain that the benefit/cost ratio is good enough to justify merging this change into Scanamo:

* **Benefit**: adds the capability to set TTL on a table through the Scanamo API, more concisely than through the AWS SDK (maybe 1-3 lines in Scanamo vs ~7 in the AWS SDK)
* **Cost**: Adds ~50 lines to the Scanamo codebase, touching 8 implementation files. (this cost would have been higher if not for the recent refactor in  https://github.com/scanamo/scanamo/pull/1786!). The code added is straightforward, but it will incrementally increase the cognitive load of managing those classes, and so have a minor ongoing maintenance cost.

Part of the issue is, given that the [DynamoDB API](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Operations_Amazon_DynamoDB.html) has 57 operations, of which Scanamo supports ~9, how much of the DynamoDB API should Scanamo attempt to cover? Do we want to multiply the size of Scanamo by 6 times to cover the entire DynamoDB API? Or is there a criteria for what operations we _would_ want to add?

For the 9 operations that Scanamo _does_ cover, the benefits of Scanamo are usually _more_ than 1 of:

* **conciseness** (which your PR does provide)
* **preventing the construction of invalid requests** (eg, `expressionAttributeNames` & `expressionAttributeValues` correctly set, avoiding reserved words, etc)
* **facilitating mapping** - between Scala/Java data types/classes and the DynamoDB item attributes model

Project code always needs maintaining, needs to be adapted and kept up to date. Scanamo is a volunteer non-commercial open-source project with only myself as a maintainer, and with this limited resource, it's valuable to me to know the benefits of every single thing in the Scanamo project are **compelling** - ie, ideally, more than 1 of the benefits above, or scoring _really highly_ on conciseness - when set against their cost.

Due to this, I don't think this PR will be merged into Scanamo, though I admit it's a fairly close thing- I do _use_ TTL on some of the projects I work on! If you have a very strong use case that justifies the inclusion in Scanamo, I'd be glad to hear it.

PR #1660, opened 2023-06-07
Add Pekko integration module

paul-danilin-moia:
Fixes https://github.com/scanamo/scanamo/issues/1655

PR #1596, opened 2022-11-02
Activate Scala 3 support

rtyley:
This takes the remainder of the Scala 3 support from @i10416's original PR: https://github.com/scanamo/scanamo/pull/1577 - all the really key Scala-3-related things are here:

* `FormatDerivation`, for abstracting the alternative approaches to Type Class Derivation - for Scala 2, we use Scala Macros, for Scala 3, we use its new automatic typeclass derivation (`scala.deriving.Mirror` etc) - see also:
  - https://docs.scala-lang.org/scala3/reference/contextual/derivation.html
  - https://www.47deg.com/blog/scala-3-deriving-typeclasses/
* Differing versions of `ScanamoCats` for Scala 2 & 3 (due to something to do with kind-projector affecting type signature, I think?)
* Activating Scala 3 support in the sbt configuration for those Scanamo modules where we can easily support Scala 3, eg the main Scanamo module, Cats Effect, Joda.
* Enabling Scala 3 in our CI build on GitHub Actions

Other preparatory parts of the PR have been squashed/tweaked slightly into these commits, already merged to the main branch:

* 36dd69ff : Update scalafmt settings prior to Scala 3
* 5453e821 & d92e3f539 : Use shared fixtures
* 0a37bf5f : Rearrange source folders, eg `scala-2.x` -> `scala-2`
* e1fd4e6e : Avoid collision with ScalaTest `===`

I'm only opening a new PR to have somewhere to easily reference all this, and because I can't easily alter the commits on @i10416's original PR as that's a forked branch I don't have control over. All credit must go to @i10416, who did a really impressive job of working out the Type Class Derivation as well as addressing the various other remaining Scala 3 based impediments to upgrading.

PR #1577, opened 2022-09-13
Support scala 3

i10416:
close https://github.com/scanamo/scanamo/issues/1465

This PR adds Scala 3 support  for core module and cats module. There are some additions/changes in build.sbt, scanamo/src/main/scala-3/org/scanamo/FormatDerivation.scala and scanamo/src/main/scala-3/org/scanamo/generic/*.

- [x] compile and test core module in scala 3
- [x] compile and test catsEffect module in scala 3
- [x] migrate tests that contains auto-derivation
- [x] move some case classes to fixtures and use share fixture models 

Following tests has not ported yet.
- scanamo/src/test/scala-w/org/scanamo/DynamoFormatTest: scala.reflect.TypeTag is not available in Scala 3
- scanamo/src/test/scala-w/org/scanamo/EnumDynamoFormatTest: it uses case class default parameter

PR #1805, opened 2024-08-26
Pekko: Support Scala 3

rtyley:
Scala 3 support was added to Scanamo with https://github.com/scanamo/scanamo/pull/1577 / https://github.com/scanamo/scanamo/pull/1596 in November 2022, but as mentioned in that 2nd PR, we only enabled Scala 3 support for Scanamo subprojects where it was nice & easy to do so - amongst others, that excluded Alpakka (the predecessor to Pekko, originally added to Scanamo with https://github.com/scanamo/scanamo/pull/151 in October 2017).

When Pekko was added to Scanamo (porting the Alpakka support) with https://github.com/scanamo/scanamo/pull/1660 in September 2023, it retained the constraint of not supporting Scala 3 - we're fixing that here.

* Scala 3 cross-compilation enabled for the Pekko project
* Pekko tests enabled under the Scala 3 CI builds
* [Fixed compilation error in `ScanamoPekko`](https://github.com/scanamo/scanamo/pull/1805/files#r1731496767) relating to Scala 3's syntax for wildcard arguments in types - we can switch to the `Pekko[A]` type alias for `Source[A, NotUsed]` and avoid specifying the unwanted type parameters anyhow. 
* [Fix compilation error in `ScanamoPekkoSpec`](https://github.com/scanamo/scanamo/pull/1805/files#r1731491275) due to changed implicit precedence in Scala 3 which led to the wrong `===` method being selected - using one from `TripleEqualsSupport` in Scalactic rather than our own Scanamo one.

### DIFF
--- a/pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
+++ b/pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
@@ -21,6 +21,7 @@ import org.apache.pekko.NotUsed
 import org.apache.pekko.actor.ClassicActorSystemProvider
 import org.apache.pekko.stream.scaladsl.{ Sink, Source }
 import org.scanamo.PekkoInstances.*
+import org.scanamo.ScanamoPekko.Pekko
 import org.scanamo.ops.{ PekkoInterpreter, ScanamoOps }
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
 
@@ -46,12 +47,12 @@ object ScanamoPekko {
 }
 
 private[scanamo] object PekkoInstances {
-  implicit val monad: Monad[Source[*, NotUsed]] = new Monad[Source[*, NotUsed]] {
-    def pure[A](x: A): Source[A, NotUsed] = Source.single(x)
+  implicit val monad: Monad[Pekko] = new Monad[Pekko] {
+    def pure[A](x: A): Pekko[A] = Source.single(x)
 
-    def flatMap[A, B](fa: Source[A, NotUsed])(f: A => Source[B, NotUsed]): Source[B, NotUsed] = fa.flatMapConcat(f)
+    def flatMap[A, B](fa: Pekko[A])(f: A => Pekko[B]): Pekko[B] = fa.flatMapConcat(f)
 
-    def tailRecM[A, B](a: A)(f: A => Source[Either[A, B], NotUsed]): Source[B, NotUsed] =
+    def tailRecM[A, B](a: A)(f: A => Pekko[Either[A, B]]): Pekko[B] =
       f(a).flatMapConcat {
         case Left(a)  => tailRecM(a)(f)
         case Right(b) => Source.single(b)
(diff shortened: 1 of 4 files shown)


pekko/src/main/scala/org/scanamo/ScanamoPekko.scala
```diff
@@ -46,7 +47,7 @@ object ScanamoPekko {
 }
 
 private[scanamo] object PekkoInstances {
-  implicit val monad: Monad[Source[*, NotUsed]] = new Monad[Source[*, NotUsed]] {
+  implicit val monad: Monad[Pekko] = new Monad[Pekko] {
```
rtyley, line 50: This fixes a compilation problem relating to [Scala 3's syntax for wildcard arguments in types](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html):

```
[error] -- [E006] Not Found Error: /home/runner/work/scanamo/scanamo/pekko/src/main/scala/org/scanamo/ScanamoPekko.scala:49:35 
[error] 49 |  implicit val monad: Monad[Source[*, NotUsed]] = new Monad[Source[*, NotUsed]] {
[error]    |                                   ^
[error]    |                                   Not found: type *
```

The [definition for `Monad`](https://github.com/typelevel/cats/blob/4e976f97bba875419854c4ad4e3f87630b7d362c/core/src/main/scala/cats/Monad.scala#L33) is `Monad[F[_]]` - we're trying to say that `Monad` takes a type `F` that has a type-parameter, but we don't care what _that_ type parameter that is (hence it's represented by the `_` underscore). We're trying to say what the monadic container type is- eg `Future`, or `List`, but we don't care what it _contains_ (eg `Future[Int]`, or `List[String]`

In the migration to Scala 3, clashing representations for both wildcards and placeholders-for-type-parameters have used `_`, `*`, _and_ `?`, as clashes between new Scala 3 syntax and the the [kind projector](https://github.com/typelevel/kind-projector) compiler plugin have been resolved.

> * In Scala 3.0, * is available as a type parameter placeholder.
> * In Scala 3.2, * is deprecated in favor of _. A -rewrite option is available to rewrite one to the other.
> * In Scala 3.3, * is removed again, and all type parameter placeholders will be expressed with _.

In the Scanamo project, we use the `-Ykind-projector:underscores` parameter for `scalac`, which gives us [_"a migration path for users that want a one-time transition to syntax with _ as a type parameter placeholder"_](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html#:~:text=a%20migration%20path,type%20parameter%20placeholder):

https://github.com/scanamo/scanamo/blob/3da03a973261f147c8668758ebfe6646b5f578f9/build.sbt#L79-L80

My _guess_ is that in the original Scanamo code (`Monad[Source[*, NotUsed]]`), `*` is being used as a placeholder-for-a-type-parameter - but in our Scala 3.3 world (with `-Ykind-projector:underscores`?), `*` isn't allowed to do that.

The fix takes advantage of the fact that Scala lets us miss out all the type-parameters anyhow, so long as...

https://github.com/scanamo/scanamo/blob/c232af6928d1ac0a5217be713d569b9442ceabbd/pekko/src/main/scala/org/scanamo/ScanamoPekko.scala#L43

PR #1804, opened 2024-08-26
Compile `refined` module under Scala 3

rtyley:
Support for `refined` was added to Scanamo with https://github.com/scanamo/scanamo/pull/182 in January 2018, but unfortunately upgrading the module to compile with Scala 3 does not work, as the tests [don't compile](https://github.com/scanamo/scanamo/actions/runs/10560310173/job/29253513792?pr=1804#step:7:487):

```
[error] -- [E007] Type Mismatch Error: /home/runner/work/scanamo/scanamo/refined/src/test/scala/org/scanamo/refined/RefinedDynamoFormatSpec.scala:19:29 
[error] 19 |    valueRead shouldBe Right(10: PosInt)
[error]    |                             ^^
[error]    |                           Found:    (10 : Int)
[error]    |                           Required: RefinedDynamoFormatSpec.this.PosInt
```

This appears to be due to https://github.com/fthomas/refined/issues/932 - implementing `refined`s macros in Scala 3 doesn't seem to be simple, and also `refineMV` has been removed, [without clear documentation on alternatives](https://github.com/fthomas/refined/issues/932#issuecomment-1291771535).

These problems with `refined` under Scala 3 do make me consider whether Scanamo's `refined` module should be removed, allowing me to simplify Scanamo's build, and reduce maintenance burden.

PR #1809, opened 2024-08-27
Remove `refined` module (lacks Scala 3 support)

rtyley:
Support for `refined` was added to Scanamo with https://github.com/scanamo/scanamo/pull/182 in January 2018, but unfortunately the [`refined` library itself has limited support for Scala 3](https://github.com/fthomas/refined/issues/932), and as seen in https://github.com/scanamo/scanamo/pull/1804, the tests for Scanamo's `refined` code do not compile under Scala 3.

Although, when [Scanamo gained Scala 3 support](https://github.com/scanamo/scanamo/pull/1596) in November 2022, we initially continued `refined` support by restricting it to release only Scala 2 modules, in order to simplify Scanamo's build and reduce ongoing maintenance burden, I'm now removing Scanamo's `refined` module.

If `refined` gains enough Scala 3 support to cross-compile against all versions of Scala supported by Scanamo, I'd definitely be happy for the `refined` module to be reintroduced to Scanamo.

## Migration path for users of Scanamo's `refined` support

If you were using the [`scanamo-refined`](https://index.scala-lang.org/scanamo/scanamo/artifacts/scanamo-refined?stable-only=true) artifact, and want to continue using the latest versions of Scanamo, you can copy this short snippet, which is the entire code of Scanamo's `refined` support:

https://github.com/scanamo/scanamo/blob/fbdc953118ed48db875b15d28dde0ddc241b1b0d/refined/src/main/scala/org/scanamo/refined/package.scala#L19-L35


build.sbt
```diff
@@ -12,8 +12,7 @@ val V = new {
   val catsEffectVersion = "3.5.4"
 }
 
-val scala2xVersions = Seq(V.scala212, V.scala213)
-val allCrossVersions = Seq(V.scala212, V.scala213, V.scala3)
+ThisBuild / scalaVersion := V.scala3
```
rtyley, line 15: This makes Scala 3 the default build for Scanamo!

